PR #18, opened 2015-04-08
Added an option to --skip-cleanup

jrjohnson:
When running in a CI environment like travis the cleanup is just extra
time wasted.  —skip-cleanup will prevent the reset from running.

I only added this to the try command, if you want I’m happy to add it to testall as well, but I'm not sure if that has the same use case.

I did a thing with a fake promise in the try task that might be the wrong approach, but I'm not super familiar with writing commands and I wasn't sure what needed to be returned from where.  Alternatively I could skip the promise and just exit, but then the output would have to be managed in two places.  
All that is to say - let me know if you want me to do this another way.

### DIFF
--- a/lib/tasks/try.js
+++ b/lib/tasks/try.js
@@ -7,9 +7,9 @@ var BowerHelpers    = require('../utils/bower-helpers');
 var findEmberPath   = require('./../utils/find-ember-path');
 
 module.exports = CoreObject.extend({
-  run: function(scenario, commandArgs){
+  run: function(scenario, commandArgs, commandOptions){
     var task = this;
-
+    var commandName = commandOptions[1] || 'test';
     process.on('SIGINT', function() {
       task.ui.writeLine( "\nGracefully shutting down from SIGINT (Ctrl-C)" );
       BowerHelpers.cleanup(task.project.root).then(function(){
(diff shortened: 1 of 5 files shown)


lib/tasks/try.js
```diff
@@ -35,7 +35,16 @@ module.exports = CoreObject.extend({
             });
         })
         .then(function(result){
-          return BowerHelpers.cleanup(task.project.root).then(function(){
+          var promise;
+          if(commandOptions.skipCleanup){
+            //create a fake promise for consistency
+            promise = new RSVP.Promise(function(resolve) {
```
rwjblue: Can just be `promise = RSVP.Promise.resolve();`.

jrjohnson:
Thanks @rwjblue I knew there was a better way to do that, but my google-foo failed me.

kategengler:
Thanks @jrjohnson. 

I think something we should warn about: If you use this option and, without cleaning up, build and deploy as the result of a passing test suite, it will build with the last set of dependencies `ember try` was run with. 

I think it would make sense for `testall` to also have this option. I expect `testall` to be the command more frequently used in CI.

jrjohnson:
@kategengler - where should I put the warning?  In the README, as output after the command is run, or in the command options itself?

kategengler:
The README would be good, thanks.

jrjohnson:
Updated the README with @kategengler's warning, added --skip-cleanup to testall, and fixed my original code to not pass that option on to ember.  Let me know if you want me to squash all these commits, I left them separate so they were easier to review.

kategengler:
Looks good, it'd be great if you could squash.

jrjohnson:
ok, all set.

PR #26, opened 2015-06-05
NPM scenario support

mike-north:
This may be quick and dirty at first, but I have a need to test  `1.13.x >= ember >= 1.7.x`, so I'm going to take a stab at it. Let's also let this be a place to start a discussion for "the proper approach" to adding NPM support. 

## Commands

No changes should be needed to the `ember try*` commands, although their internals will have to be enhanced

## Configuration - `config/ember-try.js`

```js
////////////////////////
//// EXISTING
////////////////////////

{
  name: 'ember-1.10',
  dependencies: {
    'ember': '~1.10.0'
  },
  resolutions: {
    'ember': '~1.10.0'
  }
}

////////////////////////
//// SUPPORTED
////////////////////////

{
  name: 'ember-1.10',
  dependencies: {
    'ember': '~1.10.0'
  },
  resolutions: {
    'ember': '~1.10.0'
  },
  npm: {
    dependencies: {
      'ember': '~1.10.0'
    },
    resolutions: {
      'ember': '~1.10.0'
    }
  }
}

////////////////////////
//// ENCOURAGED
////////////////////////

{
  name: 'ember-1.10',
  bower: {
    dependencies: {
      'ember': '~1.10.0'
    },
    resolutions: {
      'ember': '~1.10.0'
    },
  }
  npm: {
    dependencies: {
      'ember': '~1.10.0'
    },
    resolutions: {
      'ember': '~1.10.0'
    }
  }
}
```

## Thoughts

* The cost of a "npm install" is, for most people, way higher than the cost of a bower install.
   * It would be fantastic if adding this feature didn't increase the time it takes to run through a set of scenarios by much.
   * We could probably move the existing `node_modules` so that it's easy to put back when we want to `ember try:reset`
   * We probably do not want to blow the whole `node_modules` folder away if we don't have to
   * There's the issue of `ember-try` its self being in `node_modules`.
* For the use case of dealing w/ various versions of handlebars, some packages will need to be __removed__ from `package.json`. The current way that `bower.json` is handled supports only adding.
* I see no problem with the approach of backing up the `bower.json` file to `bower.json.ember-try`. This should apply equally well to `package.json`

- [x] Basic implementation
- [x] Save time by explicitly altering single packages (instead of a broad "npm install")
- [x] Test coverage for a variety of npm scenarios
- [x] Documentation


truenorth:
I'm still just poking around with the code for now, so it's probably not worth reviewing yet :)

mike-north:
This now works for a bunch of cases that I have tested. Just cleaning stuff up at this point, and potentially logging some more information out so people can feel confident that their scenario is being set up as they expect.

lib/utils/npm-helpers.js
```diff
@@ -0,0 +1,79 @@
+var path   = require('path');
+var fs     = require('fs-extra');
+var RSVP   = require('rsvp');
+var run    = require('./run');
+var rimraf = RSVP.denodeify(require('rimraf'));
+
+module.exports = {
+  
+  _npmFoldersToClearForScenario: function (scenario) {
```
mike-north: *NOTE TO SELF:* Needs a nice comment

lib/utils/npm-helpers.js
```diff
@@ -0,0 +1,79 @@
...
+    return ['dependencies', 'devDependencies'].reduce(function (acc, depType) {
+      var deps = (scenario.npm && scenario.npm[depType]) ? scenario.npm[depType] : [];
+      return acc.concat(Object.keys(deps));
+    }, []);
+  },
+  
+  _npmFoldersToInstallForScenario: function (scenario) {
```
mike-north: *NOTE TO SELF:* Needs a nice comment

test/helpers/mock-project.js
```diff
@@ -14,47 +16,94 @@ var readFile      = RSVP.denodeify(fs.readFile);
 
 function MockProject(root) {
   assert.equal(!!root, true, 'Must provide a root');
-  this.projectRoot = root;
+  this.projectRoot = root + '-' + Math.round(Math.random() * 10000000);
 }
 
 MockProject.prototype = {
```
mike-north: I think I will soon refactor this out into its own basic JavaScript lib, with the intent of allowing it to be a kind of Pretender for build tools and test scripts

mike-north:
Ok this works now, and is ready for review @kategengler @rwjblue

lib/tasks/testall.js
```diff
@@ -21,7 +22,10 @@ module.exports = CoreObject.extend({
           //create a fake promise for consistency
           promise = RSVP.Promise.resolve();
         } else {
-          promise = BowerHelpers.cleanup(task.project.root);
+          promise = RSVP.all([
```
kategengler: I can't comment on the exact line because github won't let me, but above there's a `backupBowerFile` call that might need to be accompanied by a `backupNpmFile`?

lib/utils/npm-helpers.js
```diff
@@ -0,0 +1,95 @@
...
+    var packages = [];
+
+    ['dependencies', 'devDependencies'].forEach(function (depType) {
+      for (pkgName in (scenario.npm[depType] || [])) {
+        if (scenario.npm[depType][pkgName]) {
+          packages.push(pkgName + "@" + scenario.npm[depType][pkgName]);
+        }
+      }
+    });
+    debugger;
```
kategengler: debugger

mike-north: yikes, sorry about that. I've now added `jshint` as part of `npm test` in addition to fixing this

https://github.com/kategengler/ember-try/commit/17ecc5966fda9de4e367a5a5c27b71a77c44fb24#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R15

mike-north:
Found a few cases where there are still problems, and I should be able to address them within the next day or two

mkorfmann:
Hey @mike-north, whats the status on this?

mike-north:
Got sidetracked on something else. Not sure when I'll be able to pick this back up, but it needs some more love before merge

ming-codes:
:+1: This would be very helpful.

Ember 1.13 loads helper automatically, but 1.12 needs another add-on to do this.

martndemus:
This would be a great feature to have merged, is there any way I can help to get this one merged?

mike-north:
it proved to be quite a bit more complicated than I initially thought (I
know, surprise!). While my initial solution works in some cases, it's quite
easy to poke holes and make it fall on its face. It's unclear when I'll
have time to give it some love

kategengler:
I've been working on npm support, but with a very different approach.

kategengler:
Closing in favor of #45

PR #28, opened 2015-06-19
Here's a test suite

truenorth:
The default ember addon testing apparatus obviously doesn't add much value for an addon like this one, so I replaced it with some mocha tests. More to come later, along with some refactoring and finally npm support

screen shot 2015-06-19 at 11 40 33 am


test/helpers/mock-project.js
```diff
@@ -0,0 +1,42 @@
...
+var fs = require('fs-extra');
+var spawnSync = require('child_process').spawnSync;
+
+var fixtureBower = require('../fixtures/bower.json');
+
+
+function MockProject(root) {
+  assert.equal(!!root, true, 'Must provide a root');
+  this.projectRoot = root;
+}
```
truenorth: The idea here is to create something like a project, so that ember-try can operate on it and then we can assert for the expected state

```
bower.json
bower_components/
```

tests/acceptance/poc-test.js
```diff
@@ -1,23 +0,0 @@
-import Ember from 'ember';
```
kategengler: I like the addition of node-side tests, I've been meaning to add them, but I would like to keep the ember tests as well. I use them as an acceptance test and for debugging ... within the project I can run `ember test` and see it run with each scenario, with my code changes without having to hunt for another project to use.

truenorth: cool, I didn't think of that. Since mocha wants `/test` and ember wants `/tests` (by default) they shouldn't conflict with each other)

lib/utils/bower-helpers.js
```diff
@@ -32,11 +32,13 @@ module.exports = {
   },
   resetBowerFile: function(root){
     var copy = RSVP.denodeify(fs.copy);
-    return copy('bower.json.ember-try', 'bower.json', {cwd: root});
```
truenorth: It doesn't appear that `cwd` is honored. I think that this was working by happenstance, since `process.cwd()` and `root` are typically identical

tests/acceptance/poc-test.js
truenorth: screen shot 2015-06-19 at 11 32 35 am
"node" and "ember" tests now run in serial

.travis.yml
```diff
@@ -3,11 +3,28 @@ language: node_js
...
   directories:
     - node_modules
 
+env:
+  - NPM_SCRIPT=node-test
+  - NPM_SCRIPT=client-test
+
+matrix:
+  allow_failures:
+    - node_js: "0.10"
```
truenorth: We use Node.js `0.10.x` internally at Yahoo, so I'd like to get this passing at some point in the future.

package.json
```diff
@@ -9,7 +9,9 @@
   "scripts": {
     "start": "ember server",
     "build": "ember build",
-    "test": "ember test"
+    "client-test": "ember test",
```
kategengler: I think if we're going to bother running this for CI, we should maybe have it be `ember try:testall` as a smoke test.

kategengler:
LGTM, save the one comment about what to run on CI. I also merged your cli upgrade first, unfortunately.

kategengler:
@truenorth This needs a rebase then I'll merge it.

kategengler:
ping @truenorth

truenorth:
@kategengler rebased

kategengler:
Thank you!

PR #31, opened 2015-06-29
Enable capturing of test stdout on testall

jnfingerle:
Fixes https://github.com/kategengler/ember-try/issues/30

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,9 @@ module.exports = {
       name: 'Ember canary',
       dependencies: {
         'ember': 'canary'
-      }
+      },
+      // Optionally capture test stdout in testall to a file
+      stdoutTestall: 'canary-test-output.tap'
     },
     {
       name: 'Ember beta',
--- a/lib/tasks/testall.js
+++ b/lib/tasks/testall.js
@@ -3,6 +3,7 @@ var CoreObject      = require('core-object');
 var RSVP            = require('rsvp');
 var mapSeries       = require('promise-map-series');
 var chalk           = require('chalk');
+var path            = require('path');
 var ScenarioManager = require('../utils/scenario-manager');
 var BowerHelpers    = require('../utils/bower-helpers');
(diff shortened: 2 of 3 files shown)


rwjblue:
I think in general, it would be better to configure `testem.json` to use a different reporter (that would make it much less prone when parsing stdout).

See https://github.com/poteto/ember-cli-flash/commit/5df7f87ab888bbebb5db6c6ff9954e7271b65a54 for an example of using `testem` output `xunit` format from test runs (IMHO this is much better than using stdout).

jnfingerle:
I'm using .tap reports that are a bit more robust for this use case. Anyway, testem doesn't have a concept of different runs with different configurations. Therefore ember-try would have to overwrite testem's report_file configuration option or provide a hook to have this file copied to another destination after each run.

rwjblue:
ember-try already modified the bower.json for each run, this could be a very similar thing. I definitely think we should avoid using stdout from 'ember test' runs, we have had MANY issues reported in ember-cli due to addons modifying the console output (stray console.logs and whatnot).

jnfingerle:
Ok, if we go this path, what should the ember-try.json file look like? Like this ...

```js
module.exports = {
  scenarios: [
    {
      name: 'Ember 1.10 with ember-data',
      dependencies: {
        'ember': '1.10.0',
        'ember-data': '1.0.0-beta.15'
      },
      testem: { /* to be merged into the existing testem.json */
        "report_file": "path/to/some/file.tap",
      }
    },
    { /*...*/ }
  ]
};
```

Or do you think of something more generic?

jnfingerle:
Ok, here's a version that changes testem.json: https://github.com/kategengler/ember-try/pull/32

mike-north:
I wonder if some more generic hooks for things like `afterScenarioSetup` and `beforeScenarioTeardown` would work for this?

jnfingerle:
For me anything that does the job is ok.

PR #32, opened 2015-07-01
Enable testem config

jnfingerle:
Fixes https://github.com/kategengler/ember-try/issues/30 - taking into account the discussion from https://github.com/kategengler/ember-try/pull/31

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ module.exports = {
       dependencies: {
         'ember': '1.10.0',
         'ember-data': '1.0.0-beta.15'
+      },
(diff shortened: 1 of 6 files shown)


rwjblue:
Take a look at https://github.com/emberjs/ember-collection/pull/90, specifically the changes to move to `testem.js` instead of `testem.json`. You can easily conditionally return different things based on environmental factors (and as of #52 the scenario is exposed as an ENV variable).

Closing this, as I believe the need is now fulfilled.  Please correct me if I have misunderstood what you needed...

PR #33, opened 2015-07-08
Adding automatically executed blueprint for config

thriqon:


rwjblue:
Thanks!  A few comments:

* The default `config/ember-try.js` should match [this](https://github.com/kategengler/ember-try/blob/master/lib/utils/config.js#L16-L44).
* We should deprecate usage without a `config/ember-try.js` present [around here](https://github.com/kategengler/ember-try/blob/master/lib/utils/config.js#L10)

thriqon:
I just updated the PR, addressing your notes, @rwjblue . Did I use `Ember.deprecate` correctly?

blueprints/ember-try/files/config/ember-try.js
```diff
@@ -0,0 +1,29 @@
+
```
rwjblue: Can you add `'use strict';` here?

lib/utils/config.js
```diff
@@ -7,6 +7,7 @@ function config(options){
     return require(configFile);
   }
   else {
+    Ember.deprecate('Relying on default configuration is deprecated', true);
```
rwjblue: We can't use `Ember.deprecate` here, because it is ran in node-land (and `Ember` global is not available).  We should use something like:

```javascript
console.warn('"config/ember-try.js" is missing from your project, please run "ember generate ember-try".');
```

kategengler:
@rwjblue Does ember-cli run the generators for addons that are installed by `ember init`? If so, should the generated `config/ember-try.js` for addons be removed from ember-cli?

Also I do not want to deprecate using ember-try without a config. See the proposal in #34

rwjblue:
> Does ember-cli run the generators for addons that are installed by ember init?

No, it does not.  If we want a `config/ember-try.js` to be generated with `ember new`, `ember addon`, or `ember init` the file has to be in the app or addon blueprints in ember-cli.

> I do not want to deprecate using ember-try without a config

Gotcha, totally your call.

kategengler:
Gotcha, thanks for the clarification. (It's already in the addon blueprints).

Turbo87:
closing due to inactivity

PR #36, opened 2015-07-20
Add support for dev dependencies

martndemus:
This commit adds support for overriding bower packages specified in `devDependencies`.

I should add (a) test(s).

### DIFF
--- a/lib/utils/scenario-manager.js
+++ b/lib/utils/scenario-manager.js
@@ -43,19 +43,31 @@ module.exports = CoreObject.extend({
   },
 
   _bowerJSONForScenario: function(bowerJSON, scenario){
-    if(!scenario.dependencies) { throw new Error("No dependencies specified for scenario " + scenario.name); }
-    var pkgs = Object.keys(scenario.dependencies);
+    if(!scenario.dependencies && !scenario.devDependencies) {
+      throw new Error("No dependencies specified for scenario " + scenario.name);
+    }
 
-    pkgs.forEach(function(pkg){
-      bowerJSON.dependencies[pkg] = scenario.dependencies[pkg];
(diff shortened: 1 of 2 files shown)


lib/utils/scenario-manager.js
```diff
@@ -47,7 +47,12 @@ module.exports = CoreObject.extend({
     var pkgs = Object.keys(scenario.dependencies);
 
     pkgs.forEach(function(pkg){
-      bowerJSON.dependencies[pkg] = scenario.dependencies[pkg];
+      if (bowerJSON.devDependencies[pkg]) {
+        bowerJSON.devDependencies[pkg] = scenario.dependencies[pkg];
```
truenorth: wouldn't you want

```js
scenario.devDependencies[pkg]
```

martndemus: I think it's not worth the distinction in this case. But I can imagine this changes somewhere in the future. If it's preferred to distinct between `dependencies` and `devDepencies` i'll change it.

truenorth: I would keep them separated, so that a scenario can be expressed as follows:

```js

{
      name: 'ember-1.11',
      dependencies: {
        'ember': '~1.11.0'
      },
      devDependencies: {
        'somePkg': '~99.99.1'
      }
}

rwjblue: @truenorth - Agreed.

@martndemus - Would you mind updating for that structure?

martndemus: @rwjblue / @truenorth sure.

martndemus:
Now even with unit tests.

rwjblue:
:+1:

truenorth:
Thanks @martndemus! this is a nice addition to the addon

PR #41, opened 2015-10-02
--allow-root flag on bower install

turboMaCk:
By default `bower install` is not working with sudo or as root. The problem is that in many automated build systems all is run with sudo by default. This adds `--allow-root` flag when running `bower install` to avoid its fail.

### DIFF
--- a/lib/utils/bower-helpers.js
+++ b/lib/utils/bower-helpers.js
@@ -27,7 +27,7 @@ module.exports = {
         return helpers.findBowerPath(root);
       })
       .then(function(bowerPath) {
-        return run('node', [bowerPath, 'install', '--config.interactive=false'], {cwd: root});
+        return run('node', [bowerPath, 'install', '--allow-root', '--config.interactive=false'], {cwd: root});
       });
   },
   resetBowerFile: function(root) {


turboMaCk:
Any feedback on this? I know it's not the best solution to allow bower to run with sudo by flag `--allow-root` but It's the best solution I came up with for now.

kategengler:
Just been pondering if we should instead expose some way of passing along whatever flags a user desires to bower.

turboMaCk:
What about prefix? `ember try:testall --bower--allow-root`

rwjblue:
We could allow custom options in `config/ember-try.js`:

```js
    {
      name: 'ember-data-2.3',
      bower: {
        commandOptions: [
          "--allow-root",
          "--other-things=whatever"
        ],
        dependencies: { },
        devDependencies: { },
        resolutions: { }
      }
    }
```

I'd much prefer that to always adding `--allow-root` here...

rwjblue:
The same could apply to `npm`...

kategengler:
I like that plan.

turboMaCk:
makes sense!

rwjblue:
Closing this PR for now, I think we'd definitely be happy with something that implements https://github.com/kategengler/ember-try/pull/41#issuecomment-177353131 if you have time...

PR #50, opened 2016-02-01
Allow --config-path to be specified to ember-try commands.

rwjblue:
This allows groups of configurations to be used with various commands.

For example, it would be nice in ember-test-helpers to group "legacy" Ember versions together and run them as a single matrix build with:

```
ember try:testall --config-path=config/legacy-scenarios.js
```

Then we could run current scenarios in another matrix build like:

```
ember try:testall --config-path=config/current-scenarios.js
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,15 @@ This addon provides a few commands:
 #### `ember try:testall`
 
 This command will run `ember test` with each scenario's specified in the config and exit appropriately.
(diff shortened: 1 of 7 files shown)


rwjblue:
@kategengler - Would love your thoughts here...

kategengler:
Sounds (and looks) good to me, thanks!

PR #52, opened 2016-02-02
Expose environment variable for currently running scenario.

rwjblue:
In some circumstances, you may want to know the scenario currently being ran (i.e. one example is generating test results files for each scenario).  This exposes the current scenario name as `EMBER_TRY_CURRENT_SCENARIO` just before running the command and resets it just before cleaning up.

My personal use case for this is to generate a testem report file for each scenario, and to do that I need to be able to detect the scenario in my projects Testem config...

See https://github.com/rwjblue/ember-collection/commit/58c88088ed8abf511f88208932c965f553c0f9c0 for an example of how I tried to work around this (and failed)...

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Can only change versions for Ember 1.10+ due to template compiler changes that t
 
 This addon provides a few commands:
 
-#### `ember try:testall`
+### `ember try:testall`
 
 This command will run `ember test` with each scenario's specified in the config and exit appropriately.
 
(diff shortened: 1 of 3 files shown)


kategengler:
This seems reasonable but I want to note that this is a little different than your workaround. Your workaround has an environment variable per config file, not per scenario. I _think_ your workaround is overwriting `SCENARIO/test-results.xml` for any config file that has more than 1 scenario.

rwjblue:
Yes, you are 100% correct, and when I realized this I came to make this PR :smiley_cat:

kategengler:
Does this warrant an example in the README of how to use the env var with Testem config?

rwjblue:
Yeah, good point.  I'll try to add in a section for this (not sure the full detail on `testem.js` and all is needed, but an explanation of "how to detect what scenario is being run" makes sense).

rwjblue:
Add a small blurb in the README, mind reviewing?

kategengler:
Thanks for adding that!

PR #53, opened 2016-02-06
Add `ember try:fastboot`

kategengler:
It installs ember-cli-fastboot, runs `ember fastboot`, then waits for the command to either exit or hit the timeout (Default 30s). Hitting the timeout means the command has succeeded in starting up a server, so this timeout may need to be adjusted longer for apps or addons with long build times.

Potential future improvements: We could separately run `ember fastboot:build` and `ember fastboot` to avoid more possible false success scenarios (when the timeout occurs before the server has started up).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # compiled output
 /dist
 /tmp
+/fastboot-dist
 
 # dependencies
 /node_modules
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ In order to use an alternate config path or to group various scenarios, you can 
 ```
 
 
+#### `ember try:fastboot`
+
+This command installs `ember-cli-fastboot` and attempts to run `ember fastboot`, to determine FastBoot compatibility.
+There is one option, `timeout` (Default: 30000), which is used to determine success. If `ember fastboot` successfully 
+starts up a sever, this command waits until the timeout to declare success.
+
 #### `ember try:reset`
 
 This command restores the original `bower.json` from `bower.json.ember-try`, `rm -rf`s `bower_components` and runs `bower install`. For use if any of the other commands fail to clean up after (they run this by default on completion).
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -1,0 +1,35 @@
+'use strict';
+
+module.exports = {
+  name: 'try:fastboot',
+  description: 'Tests whether the project is FastBoot Compatible' ,
+  works: 'insideProject',
+  availableOptions: [
+    { name: 'timeout', type: Number, default: 30000 }
+  ],
+
+  _NpmAdapter: require('../utils/npm-adapter'),
+  _TryEachTask: require('../tasks/try-each'),
+
+  run: function(commandOptions) {
+
+    var dependencyManagerAdapter = new this._NpmAdapter({cwd: this.project.root});
+    var tryEachTask = new this._TryEachTask({
+      ui: this.ui,
+      project: this.project,
+      dependencyManagerAdapters: [dependencyManagerAdapter],
+      commandArgs: ['fastboot'],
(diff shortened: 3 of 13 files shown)


rwjblue:
Much of this PR is great!

I'd rather see the ability to allow the scenario to define its own default command though.  This is now possible (since `bower` and `npm` are keys on the main scenario config):

```
module.exports = {
  command: 'ember test',  // this is the default command for all scenarios in this file, and if not present we default to `ember test`
  scenarios: [
    {
      name: 'fastboot',
      command: 'ember fastboot',  // this is the command for this individual scenario
      timeout: { time: 30000, considerTimeoutSuccess: true },
      npm: {
        'ember-cli-fastboot': '^9.9.9'
      }
    }
  ]
};
```

Making these changes makes all of our commands much better (`ember try`, `ember try:testall`, etc), and enables the functionality needed for `ember try:fastboot` without requiring the super customized command...

kategengler:
I agree re: being able to add a custom command, and that was in my first approach, but I want more time to work on that. I want to support multiple commands, and commands that aren't `ember`, which requires some refactoring. 

Since I was asked for `ember try:fastboot`, I figured something is better than nothing.

tomdale:
@kategengler One thing I've seen us do in automated tests is examine output to `stdout` and looking for messages indicating the server has started. That should significantly reduce the time required for small apps since you don't have to wait for the timeout, and means you can increase the default timeout for larger apps since hitting the timeout would be exceptional, rather than part of the normal part of testing.

(See https://github.com/tomdale/ember-cli-addon-tests/blob/master/lib/commands/start-server.js#L55-L66 for an example implementation.)

@rwjblue's plan sounds awesome but I'm not yet familiar enough with ember-try's internals to have a strong opinion. This seems fine as an MVP to me. I don't think the refactor Robert suggested affects the "public API" (`ember try:fastboot`) so there's no harm in getting this out then cleaning it up later, if I'm understanding correctly.

kategengler:
@tomdale I think @rwjblue's suggestion was to have a general way of running other commands, and so in that scenario we wouldn't have an `ember try:fastboot`. 

I thought about examining `stdout` but it seemed risky to rely on the output of another addon. If that output really won't change, then we could do that. 

Another thought I had is that maybe the `ember-cli-fastboot` addon should provide its own `ember fastboot:test` that would exit successfully if the server started up, that way it could own its own definition of what success is, and then `ember-try` can just be used to run it as a scenario ala what @rwjblue suggested (or with its own command as a convenience).

tomdale:
@kategengler The idea for `ember fastboot:test` is a great one. I can work on it but would love your help if you have spare cycles.

tomdale:
@kategengler What's the status on this? :blush:

kategengler:
@tomdale Still want to add `ember fastboot:test` to ember-cli-fastboot. Was on vacation and now blocked by https://github.com/tomdale/ember-cli-addon-tests/issues/2

tomdale:
@kategengler Cut v0.2.1 of ember-cli-addon-tests that should hopefully resolve the issue.

tomdale:
@kategengler Were you able to verify if that fixed the issue?

kategengler:
I haven't been able to get back to it yet, unfortunately.

tomdale:
@kategengler Gentle nudge. ;)

simonihmig:
Just stumbled upon this... I have put this thing together recently: https://github.com/kaliber5/ember-fastboot-addon-tests

It lets you test your addon in Fastboot. Not only spins up the server, but also lets you write custom tests that make a request and run custom assertions against the response. Also provides a `ember fastboot:test` command.

I guess this could be integrated with an ember-try scenario using the approach suggested by @rwjblue.  If there is anything else that could help making these play together nicely, I am more than willing to do my part here! :)

Turbo87:
closing due to inactivity and plenty of conflicts

PR #54, opened 2016-02-11
Refactor so that ScenarioManager doesn't need to know about config

kategengler:
Add timeout option to running a command

This is the refactor from #53 without the fastboot command, so I can get these changes sooner.


rwjblue:
This looks good to me.

PR #61, opened 2016-02-19
Add 'bowerOptions' and 'npmOptions' as top level configuration options

kategengler:
- Each are passed to their respective package manager's commands that
  are run by `ember-try`
- Both are arrays of options as strings, example: `bowerOptions:
  ['--allow-root']`


rwjblue:
LGTM - Thank you!

kategengler:
No problem!

PR #62, opened 2016-02-19
New configuration option `allowedToFail` per scenario.

### LINKED ISSUE #4: Add a parameter to the scenarios to warn on failure instead of fail
I think a useful idea would be to add a parameter to the scenarios where for a particular scenario you would warn if there was a failure instead of actually failing. This way the travis build doesn't fail if the project relies on a release ember version, when the canary ember version causes a test failure.

kategengler:
- If set, the scenario will not fail the overall `ember try` command that was run

Fixes #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,10 @@ module.exports = {
     },
     {
       name: 'Ember canary with Ember-Data 2.3.0',
+      /*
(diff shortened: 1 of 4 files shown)


rwjblue:
LGTM

PR #64, opened 2016-03-06
Some reorganization + adding debug

kategengler:
Reorganized files into folders and added debug to help debugging ember-try in the wild.


rwjblue:
LGTM

Issue #34, opened 2015-07-15
Evergreen ember release testing


lukemelia:
It would be nice to be able to specify an ember-try/travis config that tests the latest patch version of all minor ember releases above a certain specified one. As a maintainer of several plugins, it is easy for them to slip into a state where they are testing a minor version, and ember-release, but missing a minor version or two in between.

kategengler:
I agree that its tricky to keep them all up to date. 

I originally went with the scenarios because I've had things I wanted to test against multiple versions that required other packages changing in lockstep (for instance needing to go to an older ember-data along with an older ember). Something else to note is that we'll likely add npm support soon so that you could go to older versions of addons in a scenario as well. 

Given that, my first thoughts:
- Support various test reporter formats to report cumulative results or results per scenario. This will support using `ember try:testall` rather than a matrix.
- Own a concept of "allowed_failures" so that `ember try:testall` can be used instead of a build matrix. Which will allow:
- Have a "simple" mode of ember-try where it just changes ember versions, based on a semver string. It could also always do "beta" and "canary" as allowed failures by default. Currently planning to announce Ember Observer will support showing compatibility based on a key in package.json, like so:

```json
  "ember-addon": {
    "versionCompatibility": {
      "ember": "beta || canary || > 1.12.0" 
    }
  }
```

We could use this string in ember-try to test every point release available that satisfies that semver string. We would maintain the existing style of ember-try config so that anybody that needs to have multiple packages changed within a scenario could do so. 

Thoughts? Is this at all what you were imagining?

lukemelia:
This sounds like an excellent solution @kategengler.

Moving away from the matrix is unfortunate because it is so easy to see what versions are failing and what is passing, but I understand the motivation and don't see another solution.

Let me know if you want some help testing.

truenorth:
the build matrix is sort of designed to handle this sort of thing. I wonder if we could get ahold of some travis-ci folks and pick their brains about this. @stefanpenner -- any ideas?

truenorth:
Another thought -- as @lukemelia pointed out, the main value of build matrices for our purposes is having a clear indicator of passes/fails on particular versions. If we could come up with another solution that would provide similar visibility, it would be easy to let them go.

kategengler:
While the build matrix is designed to handle this sort of thing, not everybody uses travis-ci and it has the same problem @lukemelia was originally reported  -- you need to manually update with each scenario/point release you want to cover. 

[...]

PR #65, opened 2016-03-07
Add ability to auto-generate scenarios from a `versionCompatibility` statement in package.json

kategengler:
- Add command to show configuration that will be used
- versionCompatibility only works for `ember` for now

@rwjblue Ember versions are just in an array for now, but we can pull that out and get it from a package that will be maintained as you suggested. 

Open question: Do apps want to use this too? With this under `ember-addon` in package.json that's currently awkward. 

Addresses #34 

**Limitations**
- Only for Ember (for now)
- Cannot be used with travis build matrix
- Could be many versions (we will limit to latest patch version of each minor version and addons can always override with a config file if they so desire or use a different statement that will skip versions, such as `1.11.0 || 1.13.10`.

**Why not just use the versions semver statement in `package.json`?**
- The version in `package.json` is usually locked down. Also, I want addons to have to make an explicit and conscious statement of Ember versions with which they intend to be compatible.


lib/utils/auto-scenario-config-for-ember.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+var semver = require('semver');
+
+var EmberVersions = ['v2.5.0-beta.1', 'v2.4.0', 'v2.4.0-beta.3', 'v2.3.1', 'v2.4.0-beta.2', 'v2.4.0-beta.1', 'v2.3.0', 'v2.3.0-beta.3', 'v2.3.0-beta.2', 'v2.3.0-beta.1', 'v2.2.0', 'v2.1.1', 'v1.13.11', 'v2.2.0-beta.2', 'v2.2.0-beta.1', 'v2.1.0', 'v2.1.0-beta.4', 'v1.13.10', 'v2.0.2', 'v2.1.0-beta.3', 'v2.1.0-beta.2', 'v2.0.1', 'v1.13.9', 'v2.1.0-beta.1', 'v2.0.0', 'v1.13.8', 'v2.0.0-beta.5', 'v1.13.7', 'v2.0.0-beta.4', 'v1.13.6', 'v2.0.0-beta.3', 'v1.13.5', 'v1.13.4', 'v2.0.0-beta.2', 'v1.13.3', 'v1.13.2', 'v1.13.1', 'v1.13.0', 'v1.12.1', 'v1.13.0-beta.1', 'v1.12.0', 'v1.11.1', 'v1.12.0-beta.1', 'v1.11.0', 'v1.11.0-beta.5', 'v1.11.0-beta.4', 'v1.11.0-beta.3', 'v1.11.0-beta.2', 'v1.11.0-beta.1', 'v1.10.0', 'v1.10.0-beta.4', 'v1.10.0-beta.3', 'v1.10.0-beta.1', 'v1.9.1', 'v1.10.0-beta.2', 'v1.9.0', 'v1.9.0-beta.4', 'v1.9.0-beta.3', 'v1.9.0-beta.1', 'v1.8.1', 'v1.8.0', 'v1.7.1', 'v1.8.0-beta.5', 'v1.8.0-beta.4', 'v1.8.0-beta.3', 'v1.8.0-beta.2'];
```
nathanhammond: Maybe reach out to the GitHub API and look for SemVer tags?

README.md
```diff
@@ -73,6 +76,25 @@ This command is deprecated in favor of `ember try:one`. There are several bugs w
 
 ### Config
 
+##### versionCompatibility
+If you're using `ember-try` with an Ember addon, there is a short cut to test many Ember versions. In your `package.json` under the `ember-addon` key, add the following: 
+
+```json
+  "ember-addon": {
+    "versionCompatibility": {
+       "ember": ">1.11.0 <=2.0.0"
```
nathanhammond: Are there other specified patterns for this? What if we wanted to support multiple ranges? Or is that handled by YAGNI?

lib/utils/auto-scenario-config-for-ember.js
kategengler: These came from there, definitely not intended to stay this way. Github API was a thought I had but @rwjblue suggested that an npm package with version info as an alternative.

rwjblue: Github API seems ok, if it is quick and won't get blocked...

README.md
kategengler: It's anything from here https://github.com/npm/node-semver

So you can put a '> 1.11.0 || <1.3.0'. I should add a link to that package in the README, though.

nathanhammond: Oh! Brilliant! This is why I adore open source.

lib/utils/auto-scenario-config-for-ember.js
kategengler: The rate limiting might be a problem at some companies/conferences. 

The package idea is interesting, especially since we could just script updating it, but it would only be updated anytime someone `npm installed`.

nathanhammond: You could cache it for some number of hours, but that isn't likely to be respected by build machines as they blow away their local environment each time. We could also reach out to GitHub and see what their thoughts on it are.

README.md
nathanhammond: Last comment: can you extend that to support "default" "beta" and "canary"? This is close to the situation where we don't need to include an ember-try config by default.

kategengler: We chatted on slack but commenting here to document some of what we said:

"beta" and "canary" scenarios get added to whatever is generated by this statement. I like this because it requires addons to test against those channels. Those scenarios can be overridden by creating an `ember-try.js` that creates scenarios with the same name (run `ember try:config` to see what the config would have been).

PR #66, opened 2016-03-13
Make config file able to export a function

kategengler:
- Still support just an object for compatiblity
- Pass the function the project


rwjblue:
:+1:

PR #75, opened 2016-06-21
Move costly requires to be lazy.

### LINKED ISSUE #74: slow require times
noticed ember-cli is taking longer then before to boot, some of that time (20% or so for me) is spent here.

https://www.npmjs.com/package/time-require can help debug this.

some offenders appear to be:

* cli-table2
* ember-try-config

rwjblue:
Before:

```
% ember v
...snip...
        4.95 real         1.18 user         0.18 sys
```

After:

```
% ember v
...snip...
        2.09 real         1.34 user         0.27 sys
```

Fixes #74


rwjblue:
Published in 0.2.3

stefanpenner:
thanks

PR #76, opened 2016-06-21
CoreObject 1.x and 2.x compat

stefanpenner:


rwjblue:
I killed the `travis/push` job, so that the PR one would run quicker (there was a backlog of jobs in the queue).

PR #85, opened 2016-07-26
Don't use "latest" as strategy in package.json

cibernox:
"latest" as a version is a source of problems when using along with shrinkwrap, and since now ember-try is part of ember-cli, this is transitively making ember-cli a true pain to use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "core-object": "^1.1.0",
     "debug": "^2.2.0",
     "ember-cli-babel": "^5.1.3",
-    "ember-try-config": "latest",
+    "ember-try-config": "^2.0.1",
     "extend": "^3.0.0",
     "fs-extra": "^0.26.0",
     "promise-map-series": "^0.2.1",


kategengler:
I 100% agree that this addon shouldn't make ember-cli a pain to use, but changing this will also remove a feature of ember-try.

The idea was to be able to update just ember-try-config to roll out new ember versions (and other bundled scenarios in the `versionCompatbility` case) to users and they would be available on their next build run (since "latest" should always install). 

@rwjblue I'd love your thoughts on what to do here, as you gave me the idea for the ember-try-config package.

cibernox:
The (`^2.0.1`) sign allows ember-try to pick newer versions of `ember-try-config` as they are released too, as long as the version number is < 3.0.0.

kategengler:
Yes, I know. I believe that's different than `latest` though, in that if an `ember-try-config` version is already installed that satisfies that semver statement, it won't look for a newer one.

cibernox:
It is true that it won't attempt to update automatically.

I think that if there is some change that requires users to update, it is reasonable to also release a new version of ember-try with a bumped dependency on ember-try-config to `^2.0.2`. Seems to be a package that doesn't update ofter (last time was around 4months ago).

Maybe I'm missing some caveat of this approach since not ember-try is part of ember-cli.

kategengler:
It is _supposed_ to be a package that updates very often, but the feature is not yet complete. We did the separate package as a way to update the config without having to bump ember-try.

keradus:
Idea is awesome, but please don't have broken functionality exposed, as so far nobody can use shrinkwrap if he is using `ember-try` package :(

Maybe you could apply a fix and then work on strategy to achieve your goal with working shrinkwrap ?

kategengler:
Chatted with @rwjblue, we'll set an actual version as a dependency of ember-try, but then add `ember-try-config` with a version of `latest` as a dev dependency of the addon/app blueprints. Projects using shrinkwrap will then be able to remove that dependency.

cibernox:
@kategengler That should work good I think 👍

PR #96, opened 2016-09-28
Skipping deprecation warning if ember-cli is below 2.6

### LINKED ISSUE #83: Deprecation warning for Ember CLI below v2.6.0
as reported in https://github.com/ember-cli/ember-cli/issues/6058 ember-try is printing a deprecation warning even though the user project is using an Ember CLI release below v2.6.0, i.e. one that does not include ember-try.

before printing the deprecation warning we should check the `ember-cli` dependency and only print the warning if it's `2.6.0` or above.

gmurphey:
I have an old project running `ember-cli@2.5.x` that I haven't had a chance to update yet (soon!), and I was getting deprecation warnings about ember-try being included in ember-cli, but it seems that didn't happen until `ember-cli@2.6.0`. I've added a condition checking the ember-cli version around the deprecation message.

Fixes #83.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,17 @@
 /* jshint node: true */
 'use strict';
 
+var VersionChecker = require('ember-cli-version-checker');
+
 module.exports = {
   name: 'ember-try',
   init: function() {
     this._super.init && this._super.init.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

PR #98, opened 2016-09-29
Upgrade to ember-cli 2.8

kategengler:
Upgrade to ember-cli 2.8 and reduce # of travis runs to increase chances of all runs passing.


.travis.yml
```diff
@@ -14,11 +13,12 @@ cache:
...
 env:
   - NPM_SCRIPT=node-test
-  - NPM_SCRIPT=client-test
 
 matrix:
   fast_finish: true
   include:
+  - node_js: "4.2"
+    env: NPM_SCRIPT=client-test
   - node_js: "5"
```
rwjblue: Lets swap this for 6?  6 is the current stable (as it is we aren't testing current stable at all), and will become an LTS soon (technically 2016-10-01 but I think there will need to be a release first).

Issue #87, opened 2016-08-03
Allow using environment variables in command.


rwjblue:
Specifically, I would like to be able to do things like this:

``` js
    {
      name: 'ember-alpha',
      command: 'SPECIAL_SAUCE=true ember test',
      bower: {
        dependencies: {
          'ember': 'alpha'
        },
        resolutions: {
          'ember': 'alpha'
        }
      }
    }
```

or:

``` bash
ember try:one ember-alpha --- SPECIAL_SAUCE=true ember test
```

kategengler:
Fixed by #107

PR #107, opened 2016-10-26
Allow chained commands and setting env vars in commands

### LINKED ISSUE #94: Command from config cannot include multiple command with | or &&
Currently one cannot configure to run multiple commands with one liners like:

`command: 'npm uninstall X && ember test --filter unit'`

Main issue being that commands are run with child_process.spawn and not with exec. I understand that spawn is used through out the addon but maybe there is something we could do about it?

Workaround: Move command into *.sh file and set it up to be run by ember-try.

`command: "./my-script.sh"`

kategengler:
- Add note to readme about needing to npm link ember-try

Allows:

`ember try:one <scenario> --- FOO=5 ember test`
`ember try:one <scenario> --- 'ember test && echo 2'` (note quotes)

The same goes for the command in the config file. 

Fixes #94 and #87

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -20,7 +20,6 @@ bower.json
 ember-cli-build.js
 testem.js
 all-commands.sh
-test-win-commands.sh
 lint-test.js
 testem.json
 upload-coverage.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
 matrix:
   fast_finish: true
   include:
-  - node_js: "4.2"
+  - node_js: "6"
     env: NPM_SCRIPT=client-test
   - node_js: "6"
     env: NPM_SCRIPT=jscs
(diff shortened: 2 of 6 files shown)


kategengler:
@rwjblue See https://github.com/ember-cli/ember-try/pull/97 re: appveyor. Plan to at least manually check windows for now.

rwjblue:
Sounds good.

kategengler:
Manually checked on windows: All the previous types of commands seem to work, but chained commands do not.

kategengler:
Gah, not working on linux either. Works on mac, sigh.

kellyselden:
`FOO=5 ember test` is invalid syntax on Windows. I've always used `cross-env` for setting env vars multi-platform. ([here for example](https://github.com/kategengler/ember-cli-code-coverage#usage))

kategengler:
@kellyselden `cross-env` probably works already, since it's a command. Part of this work was to allow setting environment variables in the command passed to ember-try. The later part, which should be possible, isn't working on windows (chained commands with && or ||), though it might be the same issue as what I'm seeing on linux: When all-commands.sh is run, the commands use the ember-try included with ember-cli, rather than the working copy.

PR #106, opened 2016-10-14
allow removing npm packages

kellyselden:
Resolves #104

PR #108, opened 2016-10-30
Allow removing packages from bower config

kategengler:
Counterpart to #106 for bower.

As discussed, changed from 'undefined' to 'null' to signify a package should be removed in a scenario.


kategengler:
Additional commits were to get the build to pass on Travis.

PR #113, opened 2017-02-09
[Feature] Yarn

canufeel:
Allows to use yarn in `ember-try.js` config file.
- Added yarn dependency manager adapter
- Refactored dependency manager adapters to avoid code and logic duplication
- Added tests for yarn dependency manager adapter.
- Update `ember-cli` to 2.11 and other dependencies
- Use `ember-source` npm package instead of `ember` bower dependency.
- Remove node 0.12 tests from CI

Because `ember-source` npm package does not have `release`, `beta` and `canary` tags as `ember` bower package does and npm does not support tag rewriting which makes practically not possible to implement it, the question on how to deal with default `ember-try` config remains open and any suggestions are welcome.


.travis.yml
```diff
@@ -28,14 +27,14 @@ matrix:
 
 before_install:
   - npm config set spin false
-  - npm install -g bower
+  - npm install -g bower yarn
```
les2: yarn team recommends that yarn is *not* installed via npm; that said, i am currently installing via npm at work because it was hard to install OS package (ubuntu in CI). Not sure if travis makes this easier.

les2:
this is awesome. can't wait to see it land. code looks clean / much improved.

let me know if you need someone to take care of minor things like resolving the merge conflicts. i'd love to help out.

canufeel:
@les2 Hey, hi! I actually have went even further by starting to refactor tests that now seem to be a bit overwhelming in terms of loc in each file. You can view the change here - https://github.com/hanzo-es/ember-try/commit/b7e3038b70d9e9687b505e82b9cedc2c478d9a7f There are still more to go but i believe the good way would be to keep those for a separate PR as this one already seems quite a big one. I would make sure all the conflicts are resolved according to the latests changes. The thing that is currently missing is a proper documentation of current changes to guide the users from the start on how to use that feature. If you can help with those it would be great! Let me know.

.travis.yml
canufeel: Ok that's true. I would try to see if we would be able to install it via repo as it is suggested by yarn team.

les2:
i would love to help with docs.

canufeel:
@les2 That's great to hear. You can fork my repo - https://github.com/canufeel/ember-try and submit a PR to https://github.com/canufeel/ember-try/tree/feature/yarn-dep-adapter branch. 

@kategengler @rwjblue It would be great to get your feedback on what are the odds this PR is getting merged at some point and what are your views on adding `yarn` support to `ember-try`. The changes in this PR might be considered pretty big so please let us know whether you would be able to spend some time to review and merge it and whether you are ok with this PR's approach. Thanks!

kategengler:
@canufeel Thanks so much for working on this!

We have a bit of a different plan for `yarn` support: Rather than making a new adapter, `yarn` should be used in place of `npm` whenever `yarn` is installed or a `yarn.lock` is present. This will mean that most users will not have to change their existing `ember-try.js` config file in order to take advantage of `yarn`'s speedups. By default `yarn` should run with the `--no-lockfile` option. When testing different scenarios, we want users to float most dependencies to be aware of any potential problems. This is also what enables `yarn` to be used, for speed, on projects without a lockfile.  If there is a `yarn` lockfile, restoring at the end should be sure to restore it and install with yarn using the lockfile. 

We'd like to land this level of support for `yarn` ASAP. Would you be interested in working on this in a separate PR?

After that level of support lands, we'd like to add an option in the config to each scenario and top level to `useLockFile` which would in that case use the lockfile when installing with yarn.

I chatted with @rwjblue about this PR and he agrees that some of the refactors in this PR look really great, but it makes us nervous to land functionality changes, test changes and refactors all in one PR. Would you be opposed to landing changes more incrementally, in many, small PRs? It is also much easier for us to review, land, and release small, focused PRs.

canufeel:
Hey @kategengler thanks for the reply!

Everything you've said makes sense. 

I am eager to help adding `yarn` support with incremental focused PRs and i fully understand that the amount of code in this PR makes it pretty hard to review. One thing i didn't fully understand though is the following:

> Rather than making a new adapter, yarn should be used in place of npm whenever yarn is installed or a yarn.lock is present.

Is it related only to the user-faced `config/ember-try.js` syntax with its `npm` and `bower` keys for scenarios without introduction of `yarn` key or you mean that you would like to avoid adding separate `lib/dependency-manager/adadpters/yarn.js` adapter as well?

kategengler:
Great! Yes, we feel a separate adapter isn't necessary for `yarn` ... the `npm` adapter should use `yarn` to do it's business (with `--no-lockfile`) if it is installed or there is a lockfile present (in which case the end user needs to install `yarn` if it wasn't already installed).

PR #114, opened 2017-02-19
Leverage command return value instead of process.exit.

rwjblue:
Works around an upstream issue in ember-cli 2.13 canary versions, and is quite a bit easier to test.

Also tweaked SIGINT handling to queue up a cancel and avoid additional work. This is because in ember-cli 2.9+ we use `capture-exit` so that `process.exit()` does not immediately schedule an exit any longer (so the semantics that we intended with the prior SIGINT logic were generally broken already).

Addresses the issue identified in https://github.com/ember-cli/ember-cli/issues/6779.

### DIFF
--- a/lib/tasks/try-each.js
+++ b/lib/tasks/try-each.js
@@ -19,18 +19,11 @@ module.exports = CoreObject.extend({
                                                 dependencyManagerAdapters: dependencyManagerAdapters
     });
 
-    var shutdownCount = 0;
+    task._canceling = false;
     task._on('SIGINT', function() {
-      if (shutdownCount === 0) {
-        shutdownCount++;
-        task.ui.writeLine('\nGracefully shutting down from SIGINT (Ctrl-C)');
-        return task.ScenarioManager.cleanup().then(function() {
-          task._exit();
-        });
-      } else {
-        task.ui.writeLine('\nOk, but it\'s going to be a mess');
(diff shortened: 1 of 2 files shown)


lib/tasks/try-each.js
```diff
@@ -19,18 +19,11 @@ module.exports = CoreObject.extend({
                                                 dependencyManagerAdapters: dependencyManagerAdapters
     });
 
-    var shutdownCount = 0;
+    task._canceling = false;
     task._on('SIGINT', function() {
-      if (shutdownCount === 0) {
```
rwjblue: FWIW, this two phase exit system will not work on recent ember-cli anyways because `process.exit()` no longer ensures immediate exit (as it did before).

rwjblue:
Before this PR (when using ember-cli#master) a failing test suite with `ember try:one` will report `0` exit code. After this PR (when using ember-cli#master) a failing test suite with `ember try:one` will report `1` exit code.

You can test using the demo repo I created for https://github.com/ember-cli/ember-cli/issues/6779 (https://github.com/rwjblue/exit-code-incorrect-demo).

rwjblue:
FWIW, I also confirmed as far back as ember-cli@0.2.7 that returning an exit code from a command would affect the resulting process exit code, so this is not a backwards breaking change.

PR #117, opened 2017-02-27
Minimal yarn support

canufeel:
- use `yarn` package manager instead of npm if `yarn.lock` is present or `yarn --version` command returns anything other than error.
- if `yarn.lock` is present on the project it gets backed up / restored / cleaned up.

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -77,16 +77,22 @@ module.exports = CoreObject.extend({
     });
   },
   cleanup: function() {
-    var adapter = this;
-    return adapter._restoreOriginalDependencies().then(function() {
+    return this._restoreOriginalDependencies().then(() => {
       debug('Remove backup package.json and node_modules');
-      return RSVP.all([rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
-        rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))]);
+
+      let cleanupTasks = [rimraf(path.join(this.cwd, this.packageJSONBackupFileName)),
+        rimraf(path.join(this.cwd, this.nodeModulesBackupLocation))];
+
+      if (this.yarnLockPresent) {
+        cleanupTasks.push(rimraf(path.join(this.cwd, this.yarnLockBackupFileName)));
+      }
+
+      return RSVP.all(cleanupTasks);
     }).catch(function(e) {
       console.log('Error cleaning up npm scenario:', e);
     })
-    .then(function() {
-      return adapter._install();
+    .then(() => {
+      return this._install();
     });
   },
   _findCurrentVersionOf: function(packageName) {
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -16,8 +16,6 @@ module.exports = CoreObject.extend({
     this.adapterInit();
   },
   adapterInit: function() {
-    this.backupYarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLockBackupFileName));
-    this.yarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLock));
     try {
       execFileSync('yarn', [].concat(['--version']));
       this.useYarn = true;
@@ -28,9 +26,7 @@ module.exports = CoreObject.extend({
     }
   },
   useYarn: false,
-  backupYarnLockPresent: false,
   noLock: false,
-  yarnLockPresent: false,
   yarnLock: 'yarn.lock',
   yarnLockBackupFileName: 'yarn.lock.ember-try',
   configKey: 'npm',
@@ -77,13 +73,14 @@ module.exports = CoreObject.extend({
   },
   cleanup: function() {
     var adapter = this;
-    return this._restoreOriginalDependencies().then(function() {
+    var backupYarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLockBackupFileName));
+    return this._restoreOriginalDependencies(backupYarnLockPresent).then(function() {
       debug('Remove backup package.json and node_modules');
 
       var cleanupTasks = [rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
         rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))];
 
-      if (adapter.backupYarnLockPresent) {
+      if (backupYarnLockPresent) {
         cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.yarnLockBackupFileName)));
         adapter.noLock = false;
       }
@@ -147,7 +144,7 @@ module.exports = CoreObject.extend({
       }
     });
   },
-  _restoreOriginalDependencies: function() {
+  _restoreOriginalDependencies: function(backupYarnLockPresent) {
     var copy = RSVP.denodeify(fs.copy);
 
     debug('Restoring original package.json and node_modules');
@@ -159,7 +156,7 @@ module.exports = CoreObject.extend({
            path.join(this.cwd, this.nodeModules), { clobber: true })
     ];
 
-    if (this.backupYarnLockPresent) {
+    if (backupYarnLockPresent) {
       restoreTasks.push(copy(path.join(this.cwd, this.yarnLockBackupFileName),
         path.join(this.cwd, this.yarnLock)));
     }
@@ -177,10 +174,9 @@ module.exports = CoreObject.extend({
       copy(path.join(this.cwd, this.nodeModules),
            path.join(this.cwd, this.nodeModulesBackupLocation), { clobber: true })];
 
-    if (this.yarnLockPresent && this.useYarn) {
+    if (fs.existsSync(path.join(this.cwd, this.yarnLock)) && this.useYarn) {
       backupTasks.push(copy(path.join(this.cwd, this.yarnLock),
         path.join(this.cwd, this.yarnLockBackupFileName)));
-      this.backupYarnLockPresent = true;
     }
 
     return RSVP.all(backupTasks);
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -7,13 +7,37 @@ var path = require('path');
 var extend = require('extend');
 var debug = require('debug')('ember-try:dependency-manager-adapter:npm');
 var rimraf = RSVP.denodeify(require('rimraf'));
+var execFileSync = require('child_process').execFileSync;
 
 module.exports = CoreObject.extend({
   init: function() {
     this._super.apply(this, arguments);
     this.run = this.run || require('../utils/run');
+    this.adapterInit();
   },
-  configKey: 'npm',
+  adapterInit() {
+    this.yarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLock));
+    let yarnManagerPresent;
+    try {
+      execFileSync('yarn', [].concat(['--version']));
+      yarnManagerPresent = true;
+    } catch (e) {
+      yarnManagerPresent = false;
+    }
+    if (this.yarnLockPresent || yarnManagerPresent) {
+      this.useYarn = true;
+      this.configKey = 'yarn';
+      debug('Using yarn as dependency manager');
+    } else {
+      this.configKey = 'npm';
+      debug('Using npm as dependency manager');
+    }
+  },
+  useYarn: false,
+  yarnLockPresent: false,
+  yarnLock: 'yarn.lock',
+  yarnLockBackupFileName: 'yarn.lock.ember-try',
+  configKey: null,
   packageJSON: 'package.json',
   packageJSONBackupFileName: 'package.json.ember-try',
   nodeModules: 'node_modules',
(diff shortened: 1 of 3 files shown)


lib/dependency-manager-adapters/npm.js
```diff
@@ -7,13 +7,37 @@ var path = require('path');
...
     this._super.apply(this, arguments);
     this.run = this.run || require('../utils/run');
+    this.adapterInit();
   },
-  configKey: 'npm',
+  adapterInit() {
+    this.yarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLock));
+    let yarnManagerPresent;
+    try {
+      execFileSync('yarn', [].concat(['--version']));
```
rwjblue: Why use `execFileSync` here instead of [`this.run`](https://github.com/ember-cli/ember-try/blob/master/lib/utils/run.js) which returns a promise?

lib/dependency-manager-adapters/npm.js
```diff
@@ -7,13 +7,37 @@ var path = require('path');
...
+      yarnManagerPresent = true;
+    } catch (e) {
+      yarnManagerPresent = false;
+    }
+    if (this.yarnLockPresent || yarnManagerPresent) {
```
rwjblue: I vaguely think that this should only be checking `yarnManagerPresent`.  My reasoning is that even if a repo has a `yarn.lock` file, if we do not actually have access to `yarn` binary we should not attempt to use `yarn install` and whatnot).  However, if a repo does not have `yarn.lock`, but `yarn` is present we should use it (because it is massively faster than `npm install`).

lib/dependency-manager-adapters/npm.js
```diff
@@ -43,7 +67,7 @@ module.exports = CoreObject.extend({
           name: dep,
           versionExpected: deps[dep],
           versionSeen: adapter._findCurrentVersionOf(dep),
-          packageManager: 'npm'
+          packageManager: adapter.configKey
```
rwjblue: I didn't think that we would use an alternate config key in the `config/ember-try.js`, I assumed that we would continue to use `npm` as the key for each scenario...

lib/dependency-manager-adapters/npm.js
```diff
@@ -53,16 +77,22 @@ module.exports = CoreObject.extend({
...
+        cleanupTasks.push(rimraf(path.join(this.cwd, this.yarnLockBackupFileName)));
+      }
+
+      return RSVP.all(cleanupTasks);
     }).catch(function(e) {
       console.log('Error cleaning up npm scenario:', e);
     })
-    .then(function() {
-      return adapter._install();
+    .then(() => {
```
rwjblue: I cannot recall our support matrix.  I'd suspect that to rely on node@4 we would need to bump our version to 0.3.x, but I'd prefer to land these "use yarn if possible" changes in 0.2.x.

So lets stick to only node@0.12 compatible syntax for now...

lib/dependency-manager-adapters/npm.js
canufeel: `init` code is intended to run in sync. This behaviour is expected by i.e. `tasks/try-each.js` and probably by some tests and other code as well. This piece is ran in `init` because it seems the single input point for adapter. 

To make it async we can either run it in both `cleanup` and `setup` as those are async functions, but then a little bit more sophisticated logic would be required to avoid double execution or I can try to mitigate `sync` requirement for `init` but that would require works on both `bower` `dependency-manager-adapter` and probably `tasks/try-each.js` plus tests code.

lib/dependency-manager-adapters/npm.js
canufeel: That implements the logic suggested by @kategengler 

> We have a bit of a different plan for yarn support: Rather than making a new adapter, yarn should be used in place of npm whenever yarn is installed **or** a yarn.lock is present. 

I also think that access to `yarn` binary should be required. Its clear that otherwise we might hit error otherwise. However i believe that we at least should notify the user, probably via `debug` message if `yarn.lock` file was present but we are not using it because we can't find the binary.

lib/dependency-manager-adapters/npm.js
canufeel: I think that this `packageManager` key is used just for displaying versions of packages being installed in the results table which is drawn to the user after all test routines are finished. The fact that the displayed table has proper data is untested btw.

lib/dependency-manager-adapters/npm.js
canufeel: I am not familiar with that matrix but i have got an impression that `0.12` support was already dropped:
https://github.com/ember-cli/ember-try/commit/0eb4c63dc10001a042f1bf12485a4c8e4691894e#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L21

so that is a little bit confusing.

I am still ok with using older syntax and would fix it.

lib/dependency-manager-adapters/npm.js
kategengler: I suggested that it would be either/or ... I was thinking of the scenario where I add a `yarn.lock` to a project, then CI runs, but I neglected to install `yarn` on my CI system. I would expect ember-try to use yarn, but it uses npm and everything works, so I wouldn't realize it wasn't using yarn except by looking through all the output.  Maybe we should warn if using `npm` with a yarn.lock present?

lib/dependency-manager-adapters/npm.js
kategengler: `setup` is guaranteed to be called on the `dependency-manager-adapter` before anything else. Not sure why it would need to be done in cleanup as well.

kategengler:
General thought: what should we do with `npmOptions` in the case of using `yarn`? 

Also we should probably add `yarnOptions` to the config. 

@rwjblue?

lib/dependency-manager-adapters/npm.js
kategengler: @canufeel I got a little over-eager updating the travis config.

lib/dependency-manager-adapters/npm.js
canufeel: @kategengler Thinking about it twice i came out with a thought that https://github.com/ember-cli/ember-try/pull/117#discussion_r103467909 would probably stand for a proper indication of package manager being used.

lib/dependency-manager-adapters/npm.js
canufeel: what about `ember try:reset` I think it would call directly `cleanup`avoiding `setup`

kategengler: Good point, though it's likely that reset should call `setup`

lib/dependency-manager-adapters/npm.js
```diff
@@ -7,13 +7,37 @@ var path = require('path');
...
+    let yarnManagerPresent;
+    try {
+      execFileSync('yarn', [].concat(['--version']));
+      yarnManagerPresent = true;
+    } catch (e) {
+      yarnManagerPresent = false;
+    }
+    if (this.yarnLockPresent || yarnManagerPresent) {
+      this.useYarn = true;
+      this.configKey = 'yarn';
```
canufeel: We for sure can setup some other property here which would signify usage of `yarn` vs `npm`. However I am not really clear about the purpose of such change. Could you please elaborate a little bit on your thought and potential gains. Thanks in advance!

lib/dependency-manager-adapters/npm.js
```diff
@@ -53,16 +77,22 @@ module.exports = CoreObject.extend({
...
+      }
+
+      return RSVP.all(cleanupTasks);
     }).catch(function(e) {
       console.log('Error cleaning up npm scenario:', e);
     })
-    .then(function() {
-      return adapter._install();
+    .then(() => {
+      return this._install();
```
canufeel: ok noted. would fix.

lib/dependency-manager-adapters/npm.js
canufeel: @kategengler I am not really that familiar with `Eslint`. Maybe we can force `0.12` syntax with it?

lib/dependency-manager-adapters/npm.js
kategengler: The `configKey` property is used here to determine the config key used in the dependency set object https://github.com/ember-cli/ember-try/blob/master/lib/dependency-manager-adapters/npm.js#L26, which eventually originates in the `config/ember-try.js`

Some of my thoughts on support of yarn in this way (rather than through a separate adapter):
- Users with `yarn` installed get to use and benefit from `yarn install` without changing ember-try config. 
- Users with a `yarn.lock` and/or `yarn` installed (TBD elsewhere in these comments) have yarn used for their ember-try scenarios. The config stays under `npm` because even though the end user is using `yarn`, `package.json` and node_modules are the source of the dependencies, since we will ignore the `yarn.lock` (We _want_/_need_ dependencies to float for the purpose of ember-try).
- No duplication of generated configs just to support `yarn` as well, when the contents of the config will be the same. (Such as that from `ember-cli` blueprint or `ember-try-config`

If wanting to display `yarn` in the table rather than `npm` (I'm not 100% positive this is necessary, users may expect the table to match their config), at that point we could check `this.useYarn`.

canufeel: That's true. I seem to have missed that part completely. 

The only question left here is which package manager do we display in the results table. Do we always display `npm` or use `yarn` - `npm` depending on which one was used. I prefer the later option.

@rwjblue what is your point of view on that question?

lib/dependency-manager-adapters/npm.js
```diff
@@ -7,12 +7,32 @@ var path = require('path');
...
+var execFileSync = require('child_process').execFileSync;
 
 module.exports = CoreObject.extend({
   init: function() {
     this._super.apply(this, arguments);
     this.run = this.run || require('../utils/run');
+    this.adapterInit();
   },
+  adapterInit: function() {
+    this.backupYarnLockPresent = fs.existsSync(path.join(this.cwd, this.yarnLockBackupFileName));
```
kategengler: I don't think this check makes sense here ... the backup `yarn.lock` would only exist here if the task were started from a dirty state, otherwise, the backup `yarn.lock` is moved into place by the `setup` method here.

lib/dependency-manager-adapters/npm.js
```diff
@@ -36,14 +56,17 @@ module.exports = CoreObject.extend({
     debug('Write package.json with: \n', JSON.stringify(newPackageJSON));
 
     fs.writeFileSync(packageJSONFile, JSON.stringify(newPackageJSON, null, 2));
+    if (this.useYarn) {
+      this.noLock = true;
+    }
```
kategengler: Instead of a flag for `noLock`, how about passing it as an option to `_install()`

Installs when changing to a dependency set should not use the lockfile, the install on cleanup should, if present.

canufeel: Yeah i have actually thought about that. Both solutions actually seem pretty clumsy. The one i chose - because it's one more flag. The solution with argument - because when `npm` is used the argument is not utilized in any way and should be either `false` or `{}`. which is simply not used but still result in additional housekeeping and might cause confusion as `_install()` already has some kind of `options` object =) I can refactor that actually if you think it would be better but my real intent during working on this PR was keeping code changes to the minimum.

I also think that with that the amount of `if` - `else` blocks which are used to enforce some specific logic after this PR in the `npm` adapter is threateningly high and probably a good thing would be to land some kind of refactor afterwards.

lib/dependency-manager-adapters/npm.js
canufeel: Right. I thought it's better to put it here for consistency but now i see that it causes execution of code that doesn't need to be executed during setup flow. Would move to `cleanup`. Too bad this won't be that pretty =(

canufeel:
Hey @kategengler @rwjblue anyone able to review it? Please let me know if any other changes are needed to finalize this MR. Thanks!

lib/dependency-manager-adapters/npm.js
```diff
@@ -54,10 +73,19 @@ module.exports = CoreObject.extend({
...
       debug('Remove backup package.json and node_modules');
-      return RSVP.all([rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
-        rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))]);
+
+      var cleanupTasks = [rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
+        rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))];
+
+      if (backupYarnLockPresent) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.yarnLockBackupFileName)));
+        adapter.noLock = false;
```
kategengler: By the time this `noLock` flag is set to false, the reinstall of original dependencies will have been completed.

For the project that has a yarn.lock, we want to run scearios WITHOUT a lockfile, but upon cleanup use that lockfile to restore them to their normal, locked, state.

kategengler: I still think `noLock` being passed as a parameter to `_install` would make things clearer than a flag.

canufeel: @kategengler Original deps reinstall happens a few lines below - https://github.com/hanzo-es/ember-try/blob/b910cde47a62d9feb7b65628b0333e1f55a2ed1b/lib/dependency-manager-adapters/npm.js#L93.

Would refactor to parameter.

canufeel: Done. Let me know if you are ok with that approach.

canufeel:
Any news on this? 😃

kategengler:
@canufeel Thanks for your patience and work on this :)

PR #118, opened 2017-02-28
fix CI client-test cache problem

canufeel:
Remove `node_modules/ember-try` on each travis.ci run to avoid accumulation of nested folders because of caching of `node_modules` folder and repeated `npm link` / `npm link ember-try` commands being executed on each `client-test` travis scenario run.

Accumulation example can be observed here: https://travis-ci.org/ember-cli/ember-try/jobs/206107877

### DIFF
--- a/all-commands.sh
+++ b/all-commands.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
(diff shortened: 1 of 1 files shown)


rwjblue:
I also noticed this issue when working on things, thanks for fixing!

PR #124, opened 2017-04-22
Move ember-cli-babel to a devDependency.

rwjblue:
This addon does not leverage `addon` or `addon-test-support` trees, and therefore doesn't need to have `ember-cli-babel` as a runtime dep when included in other applications.

`ember-cli-babel` here is only used for the internal testing dummy app...


kategengler:
Makes sense to me. I think I moved it to dependencies originally because it told me too when running tests against the dummy app :)

rwjblue:
This came up because I'm trying to get a default ember-cli@2.13 app to ship without requiring both babel@6 and babel@5. So I've been updating addons to babel@6 that we ship by default, in this case I just didn't think it mattered because I don't think it is being used...

rwjblue:
Confirmed locally (linked into an app), no warnings or deprecations are triggered.

PR #127, opened 2017-05-23
Remove sync-exec as it is not used anymore

polarctos:
It is not used anymore and has a security issue:
https://nodesecurity.io/advisories/310


kellyselden:
I agree, doesn't look to be used. Thanks.

Issue #133, opened 2017-07-07
yarn detection isn't complete


kellyselden:
On new Travis trusty builds, they include yarn v0.17.8 by default. This breaks here
https://travis-ci.org/ember-vr/ember-aframe-camera-extras/jobs/251025550#L2614-L2629

We should check for a `yarn.lock` file as well, so we aren't accidentally running yarn.

kategengler:
Re: Checking for a `yarn.lock`

We had intentionally decided not to check for a `yarn.lock` and to use yarn if it was installed, just for the speedup. Since this seems to have caused problems we should re-eval. #125 is related. 

Adding `--ignore-engines` seems good.

kellyselden:
Probably an ignorant question: If the app has already been `npm install`ed (no `yarn.lock`), and ember-try uses a `yarn install` to do the swapping of libs, it has to do yarn's complete indexing (assumption based on no `yarn.lock` file). Would npm be faster in this case?

kategengler:
No idea, @rwjblue?

rwjblue:
For NPM <= 4 Yarn was still faster (by a fairly large margin) even when used without an existing lock file. I haven't tested with npm@5.

backspace:
Is there an elegant way around this? `ember-pouch` builds have been failing on Travis’s Trusty images because of it; it doesn’t use Yarn but that didn’t matter until it came preinstalled on Trusty. I experimented in [this PR](https://github.com/pouchdb-community/ember-pouch/pull/187) with a hideous workaround involving overriding `$PATH` to prioritise a fake `yarn` as well as with overriding Ember CLI’s built-in `ember-try` with 0.2.11, which predates Yarn support.

The older version override has worked to get the jobs passing on Trusty.

I could try a PR for skipping `yarn` unless a lockfile is present, maybe it’s as simple as changing [`_runYarnCheck()`](https://github.com/ember-cli/ember-try/blob/dfad310c1e95f64dc7a88e4d3b2d522288b0f9b1/lib/dependency-manager-adapters/npm.js#L84) to run `yarn why` or some other command that fails when there’s no lockfile?

```
yarn why v0.23.2
error No lockfile in this directory. Run `yarn install` to generate one.
➜ echo $?
1
```

kategengler:
[...]

PR #138, opened 2017-07-27
Change Yarn detection to look for lockfile

backspace:
This is a naïve fix for #133. It would seemingly work to merge, but I’m not sure if more nuance is needed, or whether the change should extend further.

1. I made use of `throw` because it was the minimal change, since that’s what `execFileSync` would do when `yarn --version` failed. I could instead strip out the exception handling and directly use the simple boolean check of `isFile()`.

2. The existing test for this doesn’t really exercise things fully, it mocks `_runYarnCheck`. Would it be worth using a true `yarn.lock` for testing?

3. What happens if a `yarn.lock` is detected but `yarn` itself isn’t present? Should the `execFileSync` perhaps remain? 🤔

Let me know your thoughts and I’m happy to improve this. Thanks for `ember-try`, it’s endlessly useful!

### DIFF
--- a/test/tasks/try-each-test.js
+++ b/test/tasks/try-each-test.js
@@ -264,10 +264,12 @@ describe('tryEach', function() {
       });
 
       writeJSONFile('package.json', fixturePackage);
+      fs.closeSync(fs.openSync('yarn.lock', 'w'));
       fs.mkdirSync('node_modules');
       writeJSONFile('bower.json', fixtureBower);
       return tryEachTask.run(config.scenarios, {}).then(function(exitCode) {
         expect(exitCode).to.equal(0, 'exits 0 when all scenarios succeed');
+        expect(output).to.include('Detected a yarn.lock file, add useYarn: true to your configuration if you want to use Yarn to install npm dependencies.');
         expect(output).to.include('Scenario first: SUCCESS');
         expect(output).to.include('Scenario second: SUCCESS');
         expect(output).to.include('Scenario with-bower-resolutions: SUCCESS');
--- a/test/tasks/try-each-test.js
+++ b/test/tasks/try-each-test.js
@@ -264,7 +264,7 @@ describe('tryEach', function() {
       });
 
       writeJSONFile('package.json', fixturePackage);
-      fs.writeFileSync('yarn.lock', 'w');
+      fs.writeFileSync('yarn.lock', '');
       fs.mkdirSync('node_modules');
       writeJSONFile('bower.json', fixtureBower);
       return tryEachTask.run(config.scenarios, {}).then(function(exitCode) {
--- a/test/tasks/try-each-test.js
+++ b/test/tasks/try-each-test.js
@@ -264,7 +264,7 @@ describe('tryEach', function() {
       });
 
       writeJSONFile('package.json', fixturePackage);
-      fs.closeSync(fs.openSync('yarn.lock', 'w'));
+      fs.writeFileSync('yarn.lock', 'w');
       fs.mkdirSync('node_modules');
       writeJSONFile('bower.json', fixtureBower);
       return tryEachTask.run(config.scenarios, {}).then(function(exitCode) {
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ If no `config/ember-try.js` file is present, the default config will be used. Th
 
 ##### A note on npm scenarios / "What about yarn?"
 
-If `yarn` is available globally on the system the `ember-try` command is run on, all npm scenarios will use `yarn` for install with the `--no-lockfile` option. 
+If you include `useYarn: true` in your `ember-try` config, all npm scenarios will use `yarn` for install with the `--no-lockfile` option.
 If the project has a `yarn.lock`, it will be used *only* on cleanup to restore the project to a clean state. 
  
 ###### But I use yarn, why `--no-lockfile`?
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -22,6 +22,9 @@ module.exports = CoreObject.extend({
   nodeModules: 'node_modules',
   nodeModulesBackupLocation: '.node_modules.ember-try',
   setup: function(options) {
+    if (!options) {
+      options = {};
+    }
     this._runYarnCheck(options.ui);
     return this._backupOriginalDependencies();
   },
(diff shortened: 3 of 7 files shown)


kategengler:
Re: 1, I think removing the exception handling would be a nice cleanup if you have the time

Re: 2, Another test with an actual file for yarn.lock is a nice-to-have

Re: 3, I think failing if yarn is not installed but yarn.lock is committed is fine, but we need to add an "out" for people who for some reason have a yarn.lock but would like to use npm. A configuration option to force use of npm or yarn, maybe. (But I wouldn't necessarily block this PR on that)

kellyselden:
Can this check for yarn installed as well as yarn.lock? That way CI servers that don't have yarn won't start breaking with this change.

kategengler:
@backspace Thanks for working on this. After thinking it through and chatting with @kellyselden and @rwjblue, we'd like to make use of yarn be only through opt-in via the config (`useYarn: true`). We'll also warn if the user has a yarn.lock and yarn on the system and does not have useYarn: true in the config (but only if there's a scenario with npm dependencies). 

Are you interested/willing/do have time to take this PR in that direction?

backspace:
I’m into replacing this with the plan you detailed. Do you think that maybe the presence of `yarn.lock` alone without `useYarn: true` should suggest `useYarn` (regardless of whether `yarn` exists as a command), since presumably if they have the file they probably want to use it and should make sure it’s installed?

I should be able to try getting into this direction later this weekend or during the upcoming week.

I did have a [mysterious CI failure](https://travis-ci.org/ember-cli/ember-try/jobs/260833733#L3579-L3580) in my work on this to date:

```
  1) tryEach with stubbed dependency manager passes along timeout options to run:
     Error: timeout of 1200ms exceeded. Ensure the done() callback is being called in this test.
```

There’s a comment within the test specifically addressing it:

```
        // With stubbed dependency manager, timing out is warning for accidentally not using the stub
```

It was confusing to me because I didn’t think I had changed anything related to using the stub. But hopefully this error will disappear as I change things around with the current plan 🤞

kategengler:
I think that approach w/ just yarn.lock is reasonable, since it'll just be a warning.

Re: CI failure, _most_ of the time that comment is true. I haven't seen that test be flaky before, but it's possible. If it doesn't go away, I'll look into it more.

backspace:
> Re: CI failure, _most_ of the time that comment is true. I haven't seen that test be flaky before, but it's possible. If it doesn't go away, I'll look into it more.

hah I totally believe it, I probably changed something I didn’t realise would affect it.

Regarding this:

> we'd like to make use of yarn be only through opt-in via the config (`useYarn: true`)

Am I correct in understanding that this is in the `ember-try.js` file, like if it were in the dummy one:

```
module.exports = {
  scenarios: [
    {
      name: 'test1',
      bower: {
        dependencies: {
          ember: '1.10.0'
        }
      }
    },
    {
      name: 'test2',
      command: 'ember test',
      npm: {
        useYarn: true,
        dependencies: {
          'ember-feature-flags': '3.0.0'
        }
      }
    }
  ]
};
```

Or is it meant to go elsewhere?

backspace:
ah, makes sense, thank you!

backspace:
I have some progress on this. Here’s a screenshot of the warning:

image

The wording can of course be adjusted. Should it be coloured differently? I see elsewhere there’s `task.ui.writeLine` used to write coloured lines out to the terminal, but I wasn’t sure how to access that in the npm adapter.

I added `useYarn: true` to the smoke test’s `ember-try` config; it seemed to make sense because it looks like the smoke test is only run on Travis, which has a Yarn installation in the `before_install`.

It feels a bit unsatisfying to have no testing around the `yarn.lock` detection when `useYarn` isn’t `true` but it also seemed weird to assert on a particular line being output to the log… 🤔

If there’s anything you think should be added or changed, let me know!

kategengler:
Re: `task.ui.writeLine` is definitely preferred over `console.log`. You'd have to pass in `ui` to the DependencyManagerAdapter. Alternatively, if you moved the yarn check to `setup`, you could pass it in just to that method from the ScenarioManager. 

Re: testing for log, I've done it elsewhere. Here is an example: https://github.com/ember-cli/ember-try/blob/master/test/tasks/try-each-test.js#L138 Alternatively, you could refactor the check into a method that returns based on whether the yarn check passed or failed, and only write the console output based on that. It would allow you to unit test the yarn check without testing the output.

Also, 😀, thanks for the work and the detailed writeup!

backspace:
Thanks for the example of log testing general ongoing input 😀 

What do you think of the Yarn check happening in `tasks/try-each.js`? I’m considering this because it seems like it really only needs to happen once, rather than for every scenario, and it probably would be best early on so it’s more noticeable. It looks like `try:one` delegates to `try:each` under the surface, so it would be covered by this too.

kategengler:
The DependencyManagerAdapter should only be created once whether running try:one or try:each. I think it's better in the npm adapter because the check should only happen when there are scenarios with npm changes.

backspace:
Okay, understood!

Here’s a screenshot of the warning with `writeLine`:

image

I updated one of the `tasks/try-each-test.js` tests to create a `yarn.lock` and added an expectation for the warning text in the output. I’ve now also updated the `README`. Maybe this is getting closer to being ready?

test/tasks/try-each-test.js
```diff
@@ -264,10 +264,12 @@ describe('tryEach', function() {
       });
 
       writeJSONFile('package.json', fixturePackage);
+      fs.closeSync(fs.openSync('yarn.lock', 'w'));
```
kategengler: Is this preferable to `fs.writeFileSync`?

kategengler:
@backspace Just had the one question in the test, otherwise LGTM and I'll merge once you reply. Thanks again for all the work & updates.

test/tasks/try-each-test.js
backspace: hah! I just copied from Stack Overflow 😳  but I guess `writeFileSync` is newer than that answer! I’ll replace with that. 😀

kategengler:
Thanks for the change! I'll merge as soon as CI passes.

backspace:
Excellent, thanks for the guidance!

mansona:
Great work everyone 🎉  I love it when I go looking for something I need and someone has put the work in 😄 

How do I get this right now? Looking at the module I'm working on it doesn't seem to depend on ember-try directly, I'm assuming that's because now it's a default dependency of ember-cli? 

If I added the the github version of ember-try to my package.json would this work or do I need to wait until there is a version released **and** ember-cli depends on that version? 

Not trying to rush anyone just curious how I can make use of this new awesome 😂

backspace:
When I was trying workarounds, I found that putting `ember-try` directly in my `package.json` did cause it to override the Ember CLI version. So you could try that, with the branch specified as `master`? At least until a version is released.

mansona:
thanks @backspace that did indeed work 👍

kellyselden:
Released in v0.2.17

PR #143, opened 2017-08-08
Remove requirement to have dependency managers.

### LINKED ISSUE #81: Allow scenarios without dependency changes
Currently, [the `ScenarioManager` checks to ensure that there are dependency managers](https://github.com/ember-cli/ember-try/blob/master/lib/utils/scenario-manager.js#L15) which are only created if you specify [some sort of dependency change](https://github.com/ember-cli/ember-try/blob/master/lib/utils/dependency-manager-adapter-factory.js#L16-L21) in your scenarios.

However, I have [a use-case](https://github.com/trentmwillis/ember-exam/issues/41) where I would like to run multiple different scenarios without changing any dependencies. This currently results in a `No dependency manager adapter` error being thrown. I can workaround this by specifying an empty dependency change (e.g., `npm: {}`), but this feels hacky.

So, I have two questions:
1. Why is there a requirement that dependency managers must be present currently?
2. Can we remove that requirement? It seems like answer should be yes since I can hack around it now anyway.

rwjblue:
Fixes #81

### DIFF
--- a/lib/utils/scenario-manager.js
+++ b/lib/utils/scenario-manager.js
@@ -3,17 +3,6 @@
 const CoreObject = require('core-object');
 
 module.exports = CoreObject.extend({
-  init() {
-    this._super.apply(this, arguments);
-    this._checkDependencyManagerAdapters();
-  },
-
-  _checkDependencyManagerAdapters() {
-    if (!this.dependencyManagerAdapters || this.dependencyManagerAdapters.length === 0) {
-      throw new Error('No dependency manager adapter');
-    }
-  },
-
   async setup() {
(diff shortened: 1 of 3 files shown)


rwjblue:
Agreed. I'll add a test for it.  

`dependencyManagerAdapters` is initialized to an empty array in [DependencyManagerAdapterFactory.generateFromConfig](https://github.com/ember-cli/ember-try/blob/2a4857dc653a7d7b05b7f4d2700da58bfc1aba08/lib/utils/dependency-manager-adapter-factory.js#L10), so I do not think we need to guard against undefined.

Issue #150, opened 2017-10-27
Scenarios missing their deps when no bower and no package-lock.json


ef4:
If your app does not have bower and does not have a package-lock.json, it's possible to get into a state where `ember-try` installs `bower` while accidentally producing a package-lock.json that *only* contains bower. 

Then apparently during scenario setup it respects the package-lock.json and you end up missing all your packages.

rwjblue:
Yes, ember-try + npm@5 is a very bad experience at the moment.

PR #151, opened 2017-10-30
Do not install bower

### LINKED ISSUE #141: ember-cli module directory emptied of files
When using a `yarn.lock` but without bower as a dependency in `package.json`, the `node_modules/ember-cli/` is emptied of files.

* Failed build with error: https://circleci.com/gh/Vestorly/torii/614
* Successful build: https://circleci.com/gh/Vestorly/torii/615 
* Only change to fix was adding bower to `package.json`: https://github.com/Vestorly/torii/pull/382/commits/43d9737b617a5a555b5d325e23fe7ad23961997f

Currently when `yarn` is detected ember-try attempts to use it for most operations. An exception is that if bower is not a dependency, `npm` is used to install that dependency. I presume there is some negative interaction between the two systems that causes the files to be deleted.

kategengler:
Use local bower if available.
If not available, use global bower.
If not available, error. 

Fixes #141, #150

### DIFF
--- a/lib/dependency-manager-adapters/bower.js
+++ b/lib/dependency-manager-adapters/bower.js
@@ -108,17 +108,31 @@ module.exports = CoreObject.extend({
   },
   _install: function() {
     var adapter = this;
-    var options = this.managerOptions || [];
-
     return rimraf(path.join(adapter.cwd, 'bower_components'))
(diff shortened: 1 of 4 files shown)


kategengler:
Covered by #157

PR #152, opened 2017-10-30
Use --no-shrinkwrap option to npm

### LINKED ISSUE #144: Ignore package-lock.json by default
We should run with `--no-shrinkwrap` by default on `npm install`, which ignores `npm-shrinkwrap.json` and `package-lock.json` and does not generate `package-lock.json`

We did this with yarn, because for most uses of `ember-try`, we want the scenarios to get the latest of the packages available and version drift is sometimes necessary for the changes made in the scenarios to even work.

kategengler:
Avoid using package-lock.json and shrinkwrap files per https://docs.npmjs.com/cli/install (and experimentation)
Fixes #144

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -108,6 +108,8 @@ module.exports = CoreObject.extend({
       if (mgrOptions.indexOf('--ignore-engines') === -1) {
         mgrOptions = mgrOptions.concat(['--ignore-engines']);
       }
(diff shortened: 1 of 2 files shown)


rwjblue:
@kategengler - Do we also need to add `--no-package-lock` to prevent it from generating/updating the `package-lock.json`?  Or does using `--no-shrink-wrap` automatically infer `--no-package-lock`?

kategengler:
@rwjblue It also behaves as if --no-package-lock

rwjblue:
Awesome, thank you

PR #156, opened 2017-11-08
Bump ember-try-config

hjdivad:
https://github.com/ember-cli/ember-try-config/pull/15 changed behaviour in 2.2.0 to set ember-source to null.

This is desired behaviour but required tweaking some tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ install:
   # Get the latest stable version of Node 0.STABLE.latest
   - ps: Install-Product node $env:nodejs_version
   # Install PhantomJS
-  - npm install phantomjs-prebuilt
+  - npm install -g phantomjs-prebuilt
   - npm install mocha-appveyor-reporter
   - npm install -g bower
   # Typical npm stuff.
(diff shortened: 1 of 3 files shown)


rwjblue:
Looks good! But CI is pretty angry at you...

hjdivad:
@rwjblue yep both this and #157 fail specifically on node 6 from the self link

hjdivad:
So the failure on this pr is unrelated to the changes.

https://travis-ci.org/ember-cli/ember-try/jobs/299392179

this ELOOP problem looks like a regression in either npm or node, on linux

rwjblue:
Can you clear CI cache? I’ve seen this get incorrectly restored from CI cache (then made again during the run, eventually hitting ELOOP$.

hjdivad:
@rwjblue will do although

@kategengler @rwjblue `npm` `2.15.5` is only coming in from ember-cli `2.8.0`, a dev dependency.

Let's just bump the ember-cli dev dependency to a newer version (at least 2.11, but probably 2.12)


thoughts?

hjdivad:
@rwjblue @kategengler any objection to bumping the ember-cli dev dep?

kategengler:
@hjdivad Not from me

PR #157, opened 2017-11-08
Implicitly add an npm dev dependency on bower

### LINKED ISSUE #141: ember-cli module directory emptied of files

### LINKED ISSUE #150: Scenarios missing their deps when no bower and no package-lock.json

hjdivad:
For scenarios that
 1. do not already have a dependency or dev dependency on bower and
 2. have bower dependencies

Also assert that
 1. npm is ran before bower
 2. if bower is run, a local bower is found

This PR is a continuation of, and supersedes, #151.
Fixes #141 
Fixes #150

### DIFF
--- a/lib/dependency-manager-adapters/bower.js
+++ b/lib/dependency-manager-adapters/bower.js
@@ -121,17 +121,6 @@ module.exports = CoreObject.extend({
     return adapter._findBowerPath(adapter.cwd).then(function(bowerPath) {
       debug('Run bower install using bower at %s', bowerPath);
       return adapter.run('node', [].concat([bowerPath], commandParts, options), { cwd: adapter.cwd });
-    }).catch(function() {
-      debug('Local bower not found');
-      return adapter._checkForGlobalBower().then(function() {
-        debug('Using global bower');
-        return adapter.run('bower', [].concat(commandParts, options), { cwd: adapter.cwd });
-      });
-    });
-  },
-  _checkForGlobalBower: function() {
-    return this.run('bower', ['-v'], { stdio: 'ignore' }).catch(function() {
-      throw new Error('Bower must be installed either locally or globally to have bower scenarios');
     });
   },
   _bowerJSONForDependencySet: function(bowerJSON, depSet) {
--- a/lib/dependency-manager-adapters/bower.js
+++ b/lib/dependency-manager-adapters/bower.js
@@ -108,17 +108,31 @@ module.exports = CoreObject.extend({
   },
   _install: function() {
     var adapter = this;
-    var options = this.managerOptions || [];
-
     return rimraf(path.join(adapter.cwd, 'bower_components'))
       .then(function() {
-        debug('Remove bower_components');
-        return adapter._findBowerPath(adapter.cwd);
-      })
-      .then(function(bowerPath) {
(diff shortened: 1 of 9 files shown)


lib/dependency-manager-adapters/bower.js
```diff
@@ -172,26 +175,13 @@ module.exports = CoreObject.extend({
...
-      })
-      .then(function(bowerPath) {
-        return '"' + path.join(bowerPath, '..', '..', 'bin', 'bower') + '"';
-      });
-  },
-  _installBower: function() {
-    var adapter = this;
-    debug('Installing bower via npm');
-    return adapter.run('npm', [].concat(['install', 'bower@^1.3.12']), { cwd: adapter.cwd });
+    return findEmberPath(adapter.cwd).then(function(emberPath) {
```
kategengler: I know you didn't add this, but I wonder if finding bower relative to ember-cli is necessary anymore?

hjdivad: @kategengler hmm good point.  So presumably the logic would be

1. run npm before bower
2. trust that bower exists in `PROJECT_ROOT/node_modules/.bin/bower` and assert otherwise

kategengler: Yeah, or we can just leave this since it works, but just thought it worth thinking about

hjdivad: @kategengler leaving it in this PR seems like the right thing to do; it'll be easier to track & revert in a separate PR if it causes issues (although I think you're right that finding it relative to ember-cli would no longer be useful).

rwjblue:
OK, one of the CI jobs (the `client-test` job) failed for:

> The job exceeded the maximum time limit for jobs, and has been terminated.

Seemed bizarre (since that job took only 12 minutes on the last merge), I restarted it...

kategengler:
Ah, most of the scenarios in client-test were only using bower, but with this PR they will each be npm installing, which could account for the huge increase in time. Might be worth rethinking client-test; I think we can remove most of the deprecated commands, at the very least.

rwjblue:
OK, I removed deprecated commands in 9f10230 (please double check me 😄). 

Hoping this shortened things up enough...

kategengler:
So did I in https://github.com/ember-cli/ember-try/pull/160

rwjblue:
Shoot, sorry. I didn’t check my GH notifications before updating...

PR #168, opened 2017-12-27
Clarify useYarn option

oskarrough:
Since this message is shown when you run `yarn test` it is not obvious which file you're supposed to add this to. Especially if you don't know ember-try.

Now it should be easier for newcomers.

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -77,7 +77,7 @@ module.exports = CoreObject.extend({
     if (!this.useYarnCommand) {
       try {
         if (fs.statSync(path.join(this.cwd, this.yarnLock)).isFile()) {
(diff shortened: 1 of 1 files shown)


lib/dependency-manager-adapters/npm.js
```diff
@@ -77,7 +77,7 @@ module.exports = CoreObject.extend({
     if (!this.useYarnCommand) {
       try {
         if (fs.statSync(path.join(this.cwd, this.yarnLock)).isFile()) {
-          ui.writeLine(chalk.yellow('Detected a yarn.lock file, add useYarn: true to your configuration if you want to use Yarn to install npm dependencies.'));
+          ui.writeLine(chalk.yellow('Detected a yarn.lock file. Add `useYarn: true` to your `ember-try.js` configuration file if you want to use Yarn to install npm dependencies.'));
```
rwjblue: Can you update to say `config/ember-try.js` (new folks may not know where the file is either)?

oskarrough: Good point, done :)

Issue #170, opened 2018-01-05
`npm prune --no-shrinkwrap` removes packages declared with `github:repo#hash` syntax


xcambar:
Since 0.2.18 and the introduction of the `--no-shrinkwrap` to the `npm prune` command (see https://github.com/ember-cli/ember-try/blob/23bc1e62845f1786026bde54eb9a53a996003d95/lib/dependency-manager-adapters/npm.js#L111-L113), **some** packages declared with the `github:org/repo#commitish` syntax are removed.

I doubt this is an `ember-try` bug, as the results vary from one platform to another and I've narrowed it down to the `npm prune --no-shrinkwap` command.

The questions at hand here are:

* Did anyone else notice this behaviour?
* The `--no-shrinkwrap` argument to `npm prune` is undocumented, is it safe for use?
* Could the same outcome be obtained differently?

FWIW, I've created a repo which test command replicates the issue: https://github.com/xcambar/npm-prune

xcambar:
See https://github.com/npm/npm/issues/19522 for the corresponding ticket in npm

kategengler:
It is a mistake/bug that `--no-shrinkwrap` is passed to prune as well as install. However, this is the first I've heard anybody mention problems with it. I believe that `npm prune` without the `--no-shrinkwrap` option should do what ember-try needs there, but that does need to be tested.

@xcambar Can you confirm that `npm prune` without `--no-shrinkwrap` does not cause you the same problem?

kategengler:
Fixed by #171

PR #171, opened 2018-01-08
Explicitly remove `--no-shrinkwrap` from the `npm prune` call

xcambar:
Outputs options in debug when running npm prune


xcambar:
1 test failed for a timeout, maybe a restart would fix it.

rwjblue:
Restarted!

lib/dependency-manager-adapters/npm.js
```diff
@@ -114,7 +114,10 @@ module.exports = CoreObject.extend({
 
     return this.run(cmd, [].concat(['install'], mgrOptions), { cwd: this.cwd }).then(function() {
       if (!adapter.useYarnCommand) {
-        debug('Run npm prune');
+        mgrOptions = mgrOptions.filter(function(opt) {
```
kategengler: Since `prune` has no documented options that cannot be replaced with Env vars, perhaps we shouldn't use mgrOptions at all when running that command. It feels like a bug that we ever did, since the example options were always things for which prune has undocumented behavior.

xcambar:
@kategengler I've applied your suggestion and removed all options to `npm prune`.

The failing Node 6 client-test in Travis is due to a missing WeakMap polyfill (see https://travis-ci.org/ember-cli/ember-try/jobs/326400719#L1660)

xcambar:
Any news from this? Are you waiting for an action on my end?

kategengler:
@xcambar I had assumed from your comment https://github.com/ember-cli/ember-try/issues/170#issuecomment-355933348 that you were still looking into it. 

If you rebase, the tests should start passing.

xcambar:
Yay, tests pass 🎉

PR #176, opened 2018-01-17
Update default config to use ember-source for release, beta, canary.

rwjblue:


kategengler:
#184 makes this unnecessary

PR #192, opened 2018-05-14
run `npm test` instead of `ember test` by default

kellyselden:
Since the default blueprint for addons now runs `ember test` instead of `ember try:each` for `npm test`, I think this should be the new default.

If someone customizes their "test" script, for example changing it to `ember exam`, their travis run is now out of sync. They would have to remember to go into ".travis.yml" and change to

```
- node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO --skip-cleanup --- npm test
```

to keep them in sync. I think we should do this for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This addon provides a few commands:
 
 ### `ember try:each`
 
-This command will run `ember test` or the configured command with each scenario's specified in the config and exit appropriately.
+This command will run `npm test` or the configured command with each scenario's specified in the config and exit appropriately.
 
 This command is especially useful to use on CI to test against multiple `ember` versions.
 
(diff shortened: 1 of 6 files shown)


Turbo87:
this looks like a breaking change to me. for someone on an older version of the addon blueprint it would recursively run `ember-try` then 🤔

kellyselden:
Good point.

kategengler:
I also agree that it is a breaking change. However, the latest release is a beta of a major version bump.

kategengler:
Do we require `npm` to be available as a global even for `yarn` users?

rwjblue:
> Do we require npm to be available as a global even for yarn users?

I think so, but we should probably use `yarn test` if `useYarn` is true.

@kellyselden - Have any time to rebase this and make that tweak?

bertdeblock:
Closing because stale, but I can reopen a new PR if this is still desired.

PR #205, opened 2018-08-29
Update deps

kategengler:


package.json
```diff
@@ -20,7 +20,7 @@
   },
   "repository": "ember-cli/ember-try",
   "engines": {
-    "node": "^4.5 || 6.* || >= 7.*"
+    "node": "6.* || >= 7.*"
```
btecu: 7 is no longer supported, it should be 6, 8 and 10: `"6.* || 8.* || >= 10.*"`

kategengler: I have a follow-on PR to update the blueprint, just fixed it enough so that tests would run.

Issue #208, opened 2018-09-10
Yarn resolutions


ryanto:
I'm working on an addon that adds functionality to Ember Data and I've found myself in a situation where the wrong version of data ends up in my final build. My ember-try config looks something like this...

```js
// ...
{
  name: 'ember-lts-2.12',
  npm: {
    devDependencies: {
      'ember-source': '~2.12.0',
      'ember-data': '~2.12.0'
    }
  }
},
// ...
{
  name: 'ember-3.1',
  npm: {
    devDependencies: {
      'ember-source': '~3.1.0',
      'ember-data': '~3.1.0'
    }
  }
},
// ...
```

And this was working great! But then I installed addon docs, which has a direct dependency on ember data: https://github.com/ember-learn/ember-cli-addon-docs/blob/v0.5.3/package.json#L48

When my Ember application is built, my build ends up with the addon docs version of Ember data. So, my tests get run against data version 2.18, regardless of what's in my try config. Yikes!

Yarn has the option to supply [resolutions](https://yarnpkg.com/en/docs/selective-version-resolutions) for overriding a dependency of a dependency. This would allow my addon to override addon docs data dependency. I tried this, and it works on the command line, but it looks like ember-try does not support the resolutions hash. This was my config...

```js
{
  name: 'ember-lts-2.12',
  npm: {
    devDependencies: {
      'ember-source': '~2.12.0',
      'ember-data': '~2.12.0'
    },
    resolutions: {
      'ember-data': '~2.12.0'
    }
  }
},
```

But the `resolutions` did not end up in my package.json when running `ember try`.

I wanted to see if you would be open to having ember-try support resolutions?

PS: I opened an issue for [addon docs](https://github.com/ember-learn/ember-cli-addon-docs/issues/242) as well.

samselikoff:
@kategengler We're running into this again. Any chance you'd support adding this to Ember Try? We might be able to work on a PR

kategengler:
ping @rwjblue for thoughts -- I think we had discussed this in person but I sadly have forgotten all details

rwjblue:
I think its probably fine to support, but instead of adding custom support for `resolutions` we should consider if we should deeply merge _any_ key with `package.json` contents...

ryanto:
I think deeply merging all keys is good. If Sam and I were to try this, is the best place to start here: https://github.com/ember-cli/ember-try/blob/master/lib/dependency-manager-adapters/npm.js

[...]

PR #231, opened 2018-11-27
Yarn resolutions

ryanto:
When running the npm adapter with yarn this PR will add any version in the depSet to yarn resolutions. This will make sure the correct version gets installed.

See #208

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -124,15 +124,43 @@ module.exports = CoreObject.extend({
     fs.writeFileSync(packageJSONFile, JSON.stringify(newPackageJSON, null, 2));
   },
   _packageJSONForDependencySet(packageJSON, depSet) {
-
     this._overridePackageJSONDependencies(packageJSON, depSet, 'dependencies');
     this._overridePackageJSONDependencies(packageJSON, depSet, 'devDependencies');
     this._overridePackageJSONDependencies(packageJSON, depSet, 'peerDependencies');
 
+    if (this.useYarnCommand) {
+      this._addYarnResolutions(packageJSON, depSet);
+    }
+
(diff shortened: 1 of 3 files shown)


lib/dependency-manager-adapters/npm.js
```diff
@@ -124,15 +124,43 @@ module.exports = CoreObject.extend({
...
     this._overridePackageJSONDependencies(packageJSON, depSet, 'peerDependencies');
 
+    if (this.useYarnCommand) {
+      this._addYarnResolutions(packageJSON, depSet);
+    }
+
     return packageJSON;
   },
+
+  _addYarnResolutions(packageJSON, depSet) {
```
ryanto: I made this its own function. I was worried about littering `_overridePackageJSONDependencies` with a bunch of `if (this.useYarnCommand)` guards.

kategengler:
@rwjblue I think automatically adding resolutions for packages in the scenario makes sense ala bower, but only if the `resolutions` key can still be in the config and override. 

@ryanto It just occurs to me -- does `resolutions` work when running `--no-lockfile`?

ryanto:
Ok great! Yah this PR has no escape hatches, which is going to cause a lot of pain. Nice catch.

I'll change this to only allow explicit resolutions as a first step.

@kategengler re: `--no-lockfile`, I think resolutions will work since the resolutions will be in package.json. I can test & verify that though.

rwjblue:
@ryanto - Had a chance to push this forward? I still think that this is important...

ryanto:
Yes! Thanks for the reminder. Let me push up the explicit resolutions code.

ryanto:
Closing this in favor of #259, which merges `resolutions` using the existing `_overridePackageJSONDependencies` function.

PR #261, opened 2019-01-14
Update smoke-test-app

kategengler:


Turbo87:
do you know what specific change fixed CI?

travisofthenorth:
@kategengler do you have any idea which change fixed the CI issue? I'm running into this error in an app now but I can't easily upgrade ember.

kategengler:
@travisofthenorth No, unfortunately I do not know for sure. Only Ember canary was failing so I would be surprised to see the failure without updating Ember. I think updating the blueprint fixed it; possibly around the targets.js.

PR #280, opened 2019-02-12
fixing npm@5 and npm@6

### LINKED ISSUE #167: npm 5 & 6 failing to switch npm deps
As seen [here](https://github.com/kellyselden/ember-data-fetch-support/pull/107), switching to npm 5 causes the ember-data overrides to not work. The dependency checker fails saying ember-data is not installed (only for certain scenarios).

https://travis-ci.org/kellyselden/ember-data-fetch-support/jobs/319855014#L1080

mansona:
Fixes #167 

I think the explanation of this is that npm now manages pruning automatically since npm@5 and we should not be doing it manually at all. 

The documentation [here](https://docs.npmjs.com/cli/prune.html) seems to be suggesting that if you run prune when you have not updated your package-lock.json it will remove the new dependency which seems to be the case here

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -83,7 +83,6 @@ module.exports = CoreObject.extend({
     }
   },
   _install() {
-    let adapter = this;
     let mgrOptions = this.managerOptions || [];
 
     debug('Run npm install with options %s', mgrOptions);
@@ -102,11 +101,7 @@ module.exports = CoreObject.extend({
       mgrOptions = mgrOptions.concat(['--no-shrinkwrap']);
     }
 
-    return this.run(cmd, [].concat(['install'], mgrOptions), { cwd: this.cwd }).then(() => {
-      if (!adapter.useYarnCommand) {
-        return adapter.run(adapter.configKey, ['prune'], { cwd: adapter.cwd });
-      }
-    });
+    return this.run(cmd, [].concat(['install'], mgrOptions), { cwd: this.cwd });
   },
   applyDependencySet(depSet) {
     depSet = depSet[this.configKey];
--- a/test/dependency-manager-adapters/npm-adapter-test.js
+++ b/test/dependency-manager-adapters/npm-adapter-test.js
@@ -52,20 +52,13 @@ describe('npmAdapter', () => {
             expect(opts).to.have.property('cwd', tmpdir);
             return RSVP.resolve();
           },
-        }, {
-          command: 'npm prune',
-          callback(command, args, opts) {
-            runCount++;
-            expect(opts).to.have.property('cwd', tmpdir);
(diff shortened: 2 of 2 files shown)


rwjblue:
Nice sleuthing!

lib/dependency-manager-adapters/npm.js
```diff
@@ -102,11 +101,7 @@ module.exports = CoreObject.extend({
       mgrOptions = mgrOptions.concat(['--no-shrinkwrap']);
     }
 
-    return this.run(cmd, [].concat(['install'], mgrOptions), { cwd: this.cwd }).then(() => {
-      if (!adapter.useYarnCommand) {
-        return adapter.run(adapter.configKey, ['prune'], { cwd: adapter.cwd });
```
rwjblue: @mansona - Should we keep this around when using npm < 5?

kellyselden: The addon blueprint still is node 6 + npm 4 I believe.

mansona: regarding the addon blueprint https://github.com/ember-cli/ember-cli/pull/8426 😏 

but to be honest I don't actually know the benefit of keeping it around for npm@4 🤔 npm@4 is such an antiquated technology that I think we should be discouraging the use of it as much as possible. Also, It's worth noting that it was released pre yarn@1.0 🙈 and I'm assuming we don't support yarn 0.28 which is the latest version from when npm@4 was last released

kellyselden: We would have to major bump here because of the existing addons testing with npm 4.

mansona: I'm 👍 with that. Integers are cheap 🎉 

Also I think this is really quite an important change because anyone trying this with the default npm on Node 8 or Node 10 will have failing tests for their addons 😞 It's not a great DX and I think a major bump is a tiny pain compared to the issues that this is causing

kategengler: Re ^ -- It's not always true. I haven't figured out specifically what causes this problem. I am certainly using the default `npm` on newer Nodes in my day-to-day and _only_ just recently encountered/recreated this problem myself with that one repo I linked you.

kategengler: If we inspect the version of node and conditionally do this, we can release it as a bugfix and those for whom this is broken (or would be -- depending on the blueprint they start with) will get the fix without having to search out and manually upgrade ember-try.

mansona: So the specific **cause** of this problem is that when you run `npm prune` and you have something installed that isn't available in the `package-lock.json` it instantly gets deleted. 

I'm going to try this fix again in a way that will actually show the issue and work in npm@4 and npm@5 👍

PR #281, opened 2019-02-12
Fixing npm@5 and npm@6 - without breaking backwards compatibility

### LINKED ISSUE #167: npm 5 & 6 failing to switch npm deps

mansona:
Fixes #167 
Closes #280 

This is an attempt to achieve the same as https://github.com/ember-cli/ember-try/pull/280 but without breaking backwards compatibility for node@4

I have been successful in making it only run npm prune with npm@4 and have also added tests that cover that issue. You will see some changes to the build scripts and this is because of slightly different behaviour between npm and yarn when it comes to installing packed `.tgz` files (which the smoke tests rely on)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,15 +39,15 @@ jobs:
     install: yarn install --no-lockfile
   - name: All Commands Smoke Test
     script:
-    - yarn client-test
+    - npm run client-test
   - name: Node 6 Smoke Test
     node_js: '6'
     script:
-    - yarn smoke-test
+    - npm run smoke-test
   - name: Node 8 Smoke Test
     node_js: '8'
     script:
-    - yarn smoke-test
+    - npm run smoke-test
   - stage: deploy
     if: tag IS present
     node_js: "10"
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,22 +32,25 @@ jobs:
   - stage: tests
     name: Node 10 With Coverage
     script:
-    - yarn lint
-    - yarn node-test-with-coverage
+      - yarn lint
+      - yarn node-test-with-coverage
   - stage: additional tests
     name: floating dependencies
     install: yarn install --no-lockfile
   - name: All Commands Smoke Test
-    script:
-    - npm run client-test
+    script: npm run client-test
   - name: Node 6 Smoke Test
     node_js: '6'
-    script:
-    - npm run smoke-test
+    script: npm run smoke-test
   - name: Node 8 Smoke Test
     node_js: '8'
-    script:
-    - npm run smoke-test
+    script: npm run smoke-test
+  - name: Node 6 Yarn Smoke Test
+    node_js: '6'
+    script: yarn smoke-test-yarn
+  - name: Node 8 Yarn Smoke Test
+    node_js: '8'
+    script: yarn smoke-test-yarn
   - stage: deploy
     if: tag IS present
     node_js: "10"
--- a/run-smoke-test.sh
+++ b/run-smoke-test.sh
@@ -5,4 +5,9 @@ rm ember-try*.tgz || true
 npm pack
 now=$(date +%s)
 mv ember-try*.tgz "ember-try-${now}.tgz"
-cd smoke-test-app && npm i && npm i --save-dev "../ember-try-${now}.tgz" && "./${1}"
+if [[ $npm_execpath =~ ^.*yarn.*$ ]]; then
+  cd smoke-test-app && yarn upgrade "ember-try@file:../ember-try-${now}.tgz" && yarn install && "./${1}"
+else
+  cd smoke-test-app && npm install --save-dev "../ember-try-${now}.tgz" && npm install && "./${1}"
+fi
+
--- a/run-smoke-test.sh
+++ b/run-smoke-test.sh
@@ -5,4 +5,4 @@ rm ember-try*.tgz || true
 npm pack
 now=$(date +%s)
 mv ember-try*.tgz "ember-try-${now}.tgz"
-cd smoke-test-app && yarn upgrade "ember-try@file:../ember-try-${now}.tgz" && yarn install && "./${1}"
+cd smoke-test-app && npm i && npm i --save-dev "../ember-try-${now}.tgz" && "./${1}"
--- a/run-smoke-test.sh
+++ b/run-smoke-test.sh
@@ -8,6 +8,9 @@ mv ember-try*.tgz "ember-try-${now}.tgz"
 if [[ $npm_execpath =~ ^.*yarn.*$ ]]; then
   cd smoke-test-app && yarn upgrade "ember-try@file:../ember-try-${now}.tgz" && yarn install && "./${1}"
 else
-  cd smoke-test-app && npm install --save-dev "../ember-try-${now}.tgz" && npm install && "./${1}"
+  if [[ $(npm -v | cut -d '.' -f 1) -lt 5 ]]; then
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -44,10 +44,18 @@ jobs:
...
+    script:
+    - npm run smoke-test-no-yarn
   - name: Node 8 Smoke Test
     node_js: '8'
     script:
     - yarn smoke-test
+  - name: Node 8 npm Smoke Test
+    node_js: '8'
+    script:
+    - yarn smoke-test-no-yarn
```
Turbo87: shouldn't this use npm instead of yarn?

smoke-test-app/config/ember-try.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  useYarn: true,
```
Turbo87: why do we declare yarn here but then refine the lockfile and replace it with package.lock?

test/fixtures/dummy-ember-try-config-with-npm-scenarios.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
       name: 'test1',
       bower: {
         dependencies: {
-          ember: '2.18.0'
+          'ember-source': '2.18.0'
```
Turbo87: this seems wrong. on bower this was not called ember-source yet.

.travis.yml
mansona: yep good catch! I missed that one

test/fixtures/dummy-ember-try-config-with-npm-scenarios.js
mansona: yes you're completely right 🙈 I was trying to fix this and I must have overlooked that it was bower

smoke-test-app/config/ember-try.js
mansona: So the changes to the lockfile were to remove the hacky use of the packed npm package that was actually defined in the lock file. Having it there causes a lot of problems for npm. You can see by my work here https://github.com/ember-cli/ember-try/pull/281/files#diff-6d3184bf759343ad27b80888330c4ba4R10 that there needs to be slightly different behaviour between how npm does this and yarn does this.

The reason why I'm adding a package-lock.json is because the bug that this PR fixes only happens if there is a package-lock in the project that ember-try is working on. To be able to test in yarn and npm effectively we need to have lock files for both environments.

I added the `useYarn` to the configs because before this PR they were doing ember-try with yarn and while they **did** work with npm I didn't want to change the behaviour of the CI suite just to add support for npm 👍

kategengler: The smoke tests were using `npm` before though -- if you don't have `useYarn:true`, it uses `npm`

run-smoke-test.sh
```diff
@@ -5,4 +5,14 @@ rm ember-try*.tgz || true
 npm pack
 now=$(date +%s)
 mv ember-try*.tgz "ember-try-${now}.tgz"
-cd smoke-test-app && yarn upgrade "ember-try@file:../ember-try-${now}.tgz" && yarn install && "./${1}"
+cd smoke-test-app
+
+if [ "$2" == "no-yarn" ]; then
+  npm i
+  npm i --save-dev "../"
```
kategengler: Why wouldn't the `npm` smoke test also use the tarball?

test/dependency-manager-adapters/npm-adapter-test.js
```diff
@@ -52,20 +52,52 @@ describe('npmAdapter', () => {
...
+            return RSVP.resolve({stdout: '4.7.1'});
+          },
         }], { allowPassthrough: false });
 
         return new NpmAdapter({
           cwd: tmpdir,
           run: stubbedRun,
         })._install().then(() => {
-          expect(runCount).to.equal(2, 'Both commands should run');
+          expect(runCount).to.equal(3, 'Both commands should run');
```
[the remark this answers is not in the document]
mansona: 👍

run-smoke-test.sh
mansona: So it seems like this was a bit of a red herring 🤔 I originally thought it was the installing via the tarball that was causing the issue but in fact it was having it missing from the lock file that was the problem. I will make it so npm also uses the tarball 👍

smoke-test-app/config/ember-try.js
mansona: I don't know why I thought the tests were using `yarn` before 🙈 I have removed this change

mansona:
I think there may have been some confusion here and I'm at the point where I'm trying to poke at CI to get it to work instead of getting to the bottom of issues 🙈 

First of all when I said `CI is using yarn` it will have been because of the fact that the initial dependencies are getting installed in the smoke test with yarn https://travis-ci.org/ember-cli/ember-try/jobs/498579579#L598 

As there are some slight differences between npm and yarn (especially in deduping), if we're using yarn to install the initial set of packages we **really** should be using yarn to install in ember-try. If we are expecting ember-try to run with npm in the smoke tests then we would expect it to run initially with npm (and remove my slightly confusing no-yarn argument to the test script)

I'll push an example of what I mean now 👍

mansona:
And now it's failing because there is a test using yarn in one of the smoke tests 😭 

We may need to rethink this 🤔 essentially the fix for the original issue is working as expected, but now it's a problem confined to how we're testing.

Essentially we need to avoid all cases where we npm install and then use yarn in the ember-try scenario, and vice versa, but I'm not really clued in on what's going on in the CI to know **why** these problems are happening

run-smoke-test.sh
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 
 set -ex
-rm ember-try*.tgz || true
-npm pack
-now=$(date +%s)
-mv ember-try*.tgz "ember-try-${now}.tgz"
-cd smoke-test-app && yarn upgrade "ember-try@file:../ember-try-${now}.tgz" && yarn install && "./${1}"
+cd smoke-test-app && npm i && npm i --save-dev "../" && "./${1}"
```
kategengler: 🤔 I think this is the cause of the CI failures. I think we still need to `npm pack` and install the tarball

kategengler: `npm install <tarball file>` is a legit thing https://docs.npmjs.com/cli/install#synopsis

mansona: I think we might need to consider doing a major version bump on this one because without this change it fails in NPM 4 for some reason: https://travis-ci.org/ember-cli/ember-try/builds/498796103?utm_source=github_status&utm_medium=notification

kategengler: Assuming it _actually_ works on npm 4, I'd be fine not running the smoke tests under it.

mansona:
@kategengler as it turns out the last issue was actually a discrepancy between older and newer npm versions as to what was being put in the package.json file. The code itself would have been working but our own version checker thingy was complaining because of differences between file paths, `file:` prefixes etc. 

the solution, in the end, was to use a specific workaround for npm versions <5. I am not too worried about this because we can remove it and all the related hacks/code when we're doing the node@6 EOL major bump 👍

mansona:
OMG it passed 🎉 😍 

Thanks so much @kategengler for helping through this process!! And for being patient with me 😂

kategengler:
Thanks again @mansona!

PR #345, opened 2019-05-18
Add dependabot config to update weekly, and update smoke-test-app too

kategengler:


rwjblue:
👍 I'm curious what the result will be for this. Will it still have separate PR's for each dep?

kategengler:
It sadly seems like it is separate PRs for each dep, just only opened weekly.

PR #409, opened 2019-10-17
Add support for using yarn.lock file

huyphamily:
Currently, ember-try is running yarn with the --no-lockfile option, which will install packages whose version may differ from the actual current yarn.lock file in the project. To get a true dependency environment during testing, this PR aims to add support for using the project's yarn.lock file when using ember-try by adding a config flag called useYarnLockFile that users can pass in to determine whether to use the project's yarn.lock file.

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -16,11 +16,16 @@ module.exports = CoreObject.extend({
   },
   useYarnCommand: false,
   yarnLock: 'yarn.lock',
+  yarnLockBackupFileName: 'yarn.lock.ember-try',
   configKey: 'npm',
   packageJSON: 'package.json',
   packageJSONBackupFileName: 'package.json.ember-try',
   nodeModules: 'node_modules',
   nodeModulesBackupLocation: '.node_modules.ember-try',
+  npmShrinkWrap: 'npm-shrinkwrap.json',
+  npmShrinkWrapBackupFileName: 'npm-shrinkwrap.json.ember-try',
+  packageLock: 'package-lock.json',
+  packageLockBackupFileName: 'package-lock.json.ember-try',
   setup(options) {
     if (!options) {
       options = {};
@@ -33,7 +38,7 @@ module.exports = CoreObject.extend({
 
     adapter.applyDependencySet(depSet);
 
-    return adapter._install().then(() => {
+    return adapter._install(depSet).then(() => {
       let deps = extend({}, depSet.dependencies || {}, depSet.devDependencies || {});
       let currentDeps = Object.keys(deps).map((dep) => {
         return {
@@ -57,6 +62,18 @@ module.exports = CoreObject.extend({
 
       let cleanupTasks = [rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
         rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))];
+
+      if (fs.existsSync(path.join(this.cwd, this.yarnLockBackupFileName))) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.yarnLockBackupFileName)));
+      }
+
+      if (fs.existsSync(path.join(this.cwd, this.npmShrinkWrapBackupFileName))) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.npmShrinkWrapBackupFileName)));
+      }
+
+      if (fs.existsSync(path.join(this.cwd, this.packageLockBackupFileName))) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.packageLockBackupFileName)));
+      }
 
       return RSVP.all(cleanupTasks);
     }).catch((e) => {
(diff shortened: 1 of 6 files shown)


lib/dependency-manager-adapters/npm.js
```diff
@@ -16,6 +16,8 @@ module.exports = CoreObject.extend({
   },
   useYarnCommand: false,
   yarnLock: 'yarn.lock',
+  yarnLockBackupFileName: 'yarn.lock.ember-try',
+  useYarnLockFile: false,
```
dnachev: I will leave it up to @rwjblue, but `ember-cli`, and consequently `ember-try` as well, supports both `npm` and `yarn`. If we are introducing `useLockfile` option, which should apply to both NPM and Yarn, we should probably handle it for both instead of making the option only valid in one case.

huyphamily: @dnachev Yea, I was contemplating whether to support both cases. I ultimately decided to support the yarn case only and named the flag useYarnLockFile instead of useLockFile. 

If we want to support using shrinkwrap, then we can add support for this in the future if there is a need? I was thinking this would be a separate flag called useNPMShrinkwrap or something along this line.

lib/dependency-manager-adapters/npm.js
```diff
@@ -58,6 +60,10 @@ module.exports = CoreObject.extend({
       let cleanupTasks = [rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
         rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))];
 
+      if (this.useYarnLockFile) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.yarnLockBackupFileName)));
```
rwjblue: This should probably be in the main definition of `cleanupTasks` (why wouldn't we want to _always_ do this?).

lib/dependency-manager-adapters/npm.js
```diff
@@ -171,6 +177,12 @@ module.exports = CoreObject.extend({
         path.join(this.cwd, this.nodeModules), { clobber: true }),
     ];
 
+    if (this.useYarnLockFile) {
+      restoreTasks.push(
```
rwjblue: I think we should always do this (not conditionally)

lib/dependency-manager-adapters/npm.js
```diff
@@ -184,6 +196,12 @@ module.exports = CoreObject.extend({
       copy(path.join(this.cwd, this.nodeModules),
         path.join(this.cwd, this.nodeModulesBackupLocation), { clobber: true })];
 
+    if (this.useYarnLockFile) {
+      backupTasks.push(
```
rwjblue: Same as above, we should always back up the lockfile when present

rwjblue:
I chatted with @kategengler, she suggested we use `buildManagerOptions` as the function name instead of my original proposal of `buildNpmOptions()`.

huyphamily:
I see...so if someone has buildManagerOptions in their config, then we use that config instead of the default. If shrinkwrap or yarn.lock files are detected, then we back up those files and restore accordingly. I can make that change.

rwjblue:
Ya, exactly

lib/dependency-manager-adapters/npm.js
```diff
@@ -82,26 +93,35 @@ module.exports = CoreObject.extend({
...
-        mgrOptions = mgrOptions.concat(['--no-lockfile']);
+      if (!Array.isArray(mgrOptions)) {
+        throw new Error('buildManagerOptions must return an array of options');
       }
-      // npm warns on incompatible engines
-      // yarn errors, not a good experience
-      if (mgrOptions.indexOf('--ignore-engines') === -1) {
-        mgrOptions = mgrOptions.concat(['--ignore-engines']);
+    } else {
+      if (this.useYarnCommand) {
```
GGuru: Do we intend to mandate the buildManagerOptions in the future? If so, we can add a deprecation here that encourages to use the manager options for these defaults.

lib/dependency-manager-adapters/bower.js
```diff
@@ -116,18 +116,29 @@ module.exports = CoreObject.extend({
...
+  _runBowerInstall(depSet) {
     let adapter = this;
     let options = this.managerOptions || [];
-    let commandParts = ['install', '--config.interactive=false'];
+    let commandParts = ['install'];
+    options.unshift('--config.interactive=false');
+
+    // buildManagerOptions overrides all default
+    if (typeof this.buildManagerOptions === 'function') {
+      options = this.buildManagerOptions(depSet);
```
rwjblue: Since the options for `bower` and `npm` are different we need to indicate _somehow_ to `buildManagerOptions` which thing we want options for.

I see two options:

1. pass two arguments (so the invocation becomes `this.buildManagerOptions('bower', depSet)`)
2. Don't support `buildManagerOptions` with bower

I personally think we should go with 2) above. Bower itself is long deprecated, and ember-try is likely to remove support for it in the next major bump; it seems silly to make a "less ergonomic API" for a new feature just to support deprecated/legacy functionality.

lib/dependency-manager-adapters/npm.js
rwjblue: Do we really need to ensure it exists first? Doesn't `rimraf('some-path-that-doesnt-exist.txt')` resolve successfully?

lib/dependency-manager-adapters/npm.js
```diff
@@ -58,6 +61,14 @@ module.exports = CoreObject.extend({
       let cleanupTasks = [rimraf(path.join(adapter.cwd, adapter.packageJSONBackupFileName)),
         rimraf(path.join(adapter.cwd, adapter.nodeModulesBackupLocation))];
 
+      if (fs.existsSync(path.join(this.cwd, this.yarnLockBackupFileName))) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.yarnLockBackupFileName)));
+      }
+
+      if (fs.existsSync(path.join(this.cwd, this.NPMShrinkWrapBackupFileName))) {
+        cleanupTasks.push(rimraf(path.join(adapter.cwd, adapter.NPMShrinkWrapBackupFileName)));
```
rwjblue: Same as above, I would think that we can always do the `rimraf`...

lib/dependency-manager-adapters/npm.js
rwjblue: > Do we intend to mandate the buildManagerOptions in the future?

No, I believe that it will always be a "power user" option. I would expect that **most** users would use whatever our default command line options are, and you'd only specify `buildManagerOptions` if you had to deviate from that for some reason.

lib/dependency-manager-adapters/npm.js
huyphamily: @rwjblue rimraf will throw an error if the file does not exist.  We will need to check for its' existence before removing it.

lib/dependency-manager-adapters/bower.js
huyphamily: Okay, I will not be supporting bower.

lib/dependency-manager-adapters/npm.js
huyphamily: rimraf will throw an error when a file does not exist.

lib/dependency-manager-adapters/bower.js
huyphamily: @rwjblue I removed bower support.

lib/dependency-manager-adapters/npm.js
rwjblue: Ugh, ok, sorry.

lib/dependency-manager-adapters/npm.js
```diff
@@ -16,11 +16,14 @@ module.exports = CoreObject.extend({
   },
   useYarnCommand: false,
   yarnLock: 'yarn.lock',
+  yarnLockBackupFileName: 'yarn.lock.ember-try',
   configKey: 'npm',
   packageJSON: 'package.json',
   packageJSONBackupFileName: 'package.json.ember-try',
   nodeModules: 'node_modules',
   nodeModulesBackupLocation: '.node_modules.ember-try',
+  NPMShrinkWrap: 'npm-shrinkwrap.json',
```
kategengler: nit: `npm` is lowercase

lib/dependency-manager-adapters/npm.js
```diff
@@ -16,11 +16,14 @@ module.exports = CoreObject.extend({
...
+  NPMShrinkWrapBackupFileName: 'npm-shrinkwrap.json.ember-try',
```
kategengler: What about package-lock.json?

rwjblue: Ya, good catch! We should definitely retrieve and preserve `package-lock.json` in the same way that we do `yarn.lock`.

lib/dependency-manager-adapters/npm.js
huyphamily: I agree; it is backward compatible and users do not need to use it if there isn't a need.

huyphamily:
@kategengler @rwjblue Addressed all comments and added buildManagerOptions to the readme.

rwjblue:
Released in [v1.3.0](https://github.com/ember-cli/ember-try/releases/tag/v1.3.0).

PR #451, opened 2020-02-10
Restore backup files before changing to new dependency set for new scenarios.

huyphamily:
This PR is created to fix an issue we are having when using `buildManagerOptions` and removing the default `--no-lockfile` option in order to use our yarn.lock file.  In the current behavior, when switching between scenarios, versions from the previous scenario gets carried over to the next scenario, which is not the ideal behavior that we want. We want to reset the yarn.lock to the original version, before changing the dependencies of the new scenario.

In order to accomplish this, we will restore all backups before installing new dependencies for the new scenario. This will not affect the current default behavior for those with default settings, but will greatly improve the experience of those who are using `buildManagerOptions` to override the default `--no-lockfile`.

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -59,7 +59,7 @@ module.exports = CoreObject.extend({
 
   async cleanup() {
     try {
-      await this._restoreOriginalDependencies();
+      await this.restoreOriginalDependencies();
 
       debug('Remove backup package.json and node_modules');
 
@@ -204,7 +204,7 @@ module.exports = CoreObject.extend({
     });
   },
 
-  _restoreOriginalDependencies() {
+  restoreOriginalDependencies() {
     debug('Restoring original package.json and node_modules');
 
     let restoreTasks = [
--- a/lib/dependency-manager-adapters/workspace.js
+++ b/lib/dependency-manager-adapters/workspace.js
@@ -85,6 +85,12 @@ module.exports = CoreObject.extend({
     return currentDeps;
   },
 
+  restoreOriginalDependencies() {
+    return Promise.all(
+      this._packageAdapters.map((adapter) => adapter.restoreOriginalDependencies())
+    );
+  },
+
   cleanup() {
     return Promise.all(this._packageAdapters.map((adapter) => adapter.cleanup()));
   },
--- a/test/dependency-manager-adapters/npm-adapter-test.js
+++ b/test/dependency-manager-adapters/npm-adapter-test.js
@@ -354,13 +354,13 @@ describe('npmAdapter', () => {
     });
   });
 
-  describe('#_restoreOriginalDependencies', () => {
+  describe('#restoreOriginalDependencies', () => {
     it('replaces the package.json with the backed up version', () => {
       writeJSONFile('package.json.ember-try', { originalPackageJSON: true });
       writeJSONFile('package.json', { originalPackageJSON: false });
       fs.mkdirSync('.node_modules.ember-try');
       writeJSONFile('.node_modules.ember-try/prove-it.json', { originalNodeModules: true });
-      return new NpmAdapter({ cwd: tmpdir })._restoreOriginalDependencies().then(() => {
+      return new NpmAdapter({ cwd: tmpdir }).restoreOriginalDependencies().then(() => {
         assertFileContainsJSON(path.join(tmpdir, 'package.json'), { originalPackageJSON: true });
         assertFileContainsJSON(path.join(tmpdir, 'node_modules/prove-it.json'), {
           originalNodeModules: true,
@@ -379,7 +379,7 @@ describe('npmAdapter', () => {
       writeJSONFile('npm-shrinkwrap.json', { originalNpmShrinkWrap: false });
       writeJSONFile('package-lock.json.ember-try', { originalPackageLock: true });
       writeJSONFile('package-lock.json', { originalPackageLock: false });
-      return new NpmAdapter({ cwd: tmpdir })._restoreOriginalDependencies().then(() => {
+      return new NpmAdapter({ cwd: tmpdir }).restoreOriginalDependencies().then(() => {
         assertFileContainsJSON(path.join(tmpdir, 'package.json'), { originalPackageJSON: true });
         assertFileContainsJSON(path.join(tmpdir, 'node_modules/prove-it.json'), {
           originalNodeModules: true,
--- a/lib/tasks/try-each.js
+++ b/lib/tasks/try-each.js
@@ -97,6 +97,8 @@ module.exports = CoreObject.extend({
     runResults.result = result;
     this._writeFooter(`Result: ${result}`);
 
+    await this.ScenarioManager.restoreOriginalDependencies();
+
     return runResults;
   },
 
--- a/lib/tasks/try-each.js
+++ b/lib/tasks/try-each.js
@@ -40,6 +40,10 @@ module.exports = CoreObject.extend({
       let results = [];
       for (let scenario of scenarios) {
         results.push(await this._runCommandForThisScenario(scenario));
+
+        if (scenarios.length > 1) {
+          await this.ScenarioManager.restoreOriginalDependencies();
+        }
       }
 
       await this._optionallyCleanup(options);
@@ -96,8 +100,6 @@ module.exports = CoreObject.extend({
 
     runResults.result = result;
     this._writeFooter(`Result: ${result}`);
-
-    await this.ScenarioManager.restoreOriginalDependencies();
 
     return runResults;
   },
--- a/lib/utils/scenario-manager.js
+++ b/lib/utils/scenario-manager.js
@@ -41,4 +41,10 @@ module.exports = CoreObject.extend({
       await depManager.cleanup();
     }
   },
+
+  async restoreOriginalDependencies() {
+    for (let depManager of this.dependencyManagerAdapters) {
+      await depManager.restoreOriginalDependencies();
+    }
+  },
 });
--- a/test/dependency-manager-adapters/workspace-adapter-test.js
+++ b/test/dependency-manager-adapters/workspace-adapter-test.js
@@ -242,6 +242,77 @@ describe('workspaceAdapter', () => {
     });
   });
 
+  describe('#restoreOriginalDependencies', () => {
+    it('replaces the package.json with the backed up version', () => {
+      fs.ensureDirSync('packages/test/node_modules');
+      writeJSONFile('packages/test/package.json', { originalPackageJSON: false });
+      writeJSONFile('packages/test/node_modules/prove-it.json', { originalNodeModules: false });
+
+      fs.ensureDirSync('packages/test/.node_modules.ember-try');
+      writeJSONFile('packages/test/package.json.ember-try', { originalPackageJSON: true });
+      writeJSONFile('packages/test/.node_modules.ember-try/prove-it.json', {
+        originalNodeModules: true,
+      });
+
+      let workspaceAdapter = new WorkspaceAdapter({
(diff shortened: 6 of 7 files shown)


lib/dependency-manager-adapters/npm.js
```diff
@@ -147,14 +147,16 @@ module.exports = CoreObject.extend({
     debug('Changing to dependency set: %s', JSON.stringify(depSet));
 
     if (!depSet) { return RSVP.resolve([]); }
-    let backupPackageJSON = path.join(this.cwd, this.packageJSONBackupFileName);
-    let packageJSONFile = path.join(this.cwd, this.packageJSON);
-    let packageJSON = JSON.parse(fs.readFileSync(backupPackageJSON));
-    let newPackageJSON = this._packageJSONForDependencySet(packageJSON, depSet);
 
-    debug('Write package.json with: \n', JSON.stringify(newPackageJSON));
+    return this._restoreOriginalDependencies().then(() => {
```
rwjblue: It seems very bizarre to restore before applying the depSet changes here.

Any cleanup/restore that was supposed to happen should have happened **before** the next scenario runs it's `applyDependencySet`.

lib/dependency-manager-adapters/npm.js
```diff
@@ -36,22 +36,22 @@ module.exports = CoreObject.extend({
...
-    return adapter._install(depSet).then(() => {
-      let deps = extend({}, depSet.dependencies || {}, depSet.devDependencies || {});
-      let currentDeps = Object.keys(deps).map((dep) => {
-        return {
-          name: dep,
-          versionExpected: deps[dep],
-          versionSeen: adapter._findCurrentVersionOf(dep),
-          packageManager: adapter.useYarnCommand ? 'yarn' : 'npm',
-        };
+    return adapter.applyDependencySet(depSet).then(() => {
```
rwjblue: We absolutely should have been entangling this all along, this fix seems good regardless of the changes within `applyDependencySet` itself.

lib/dependency-manager-adapters/workspace.js
```diff
@@ -56,24 +56,26 @@ module.exports = CoreObject.extend({
...
-          name: dep,
-          versionExpected: deps[dep],
-          versionSeen: this._findCurrentVersionOf(dep),
-          packageManager: 'yarn',
-        };
+    let appliedDependencies = this._packageAdapters.map(adapter =>
+      adapter.applyDependencySet(depSet)
+    );
+
+    return RSVP.all(appliedDependencies).then(() => {
```
rwjblue: Like above, this seems like a great fundamental fix. Are you certain that these fixes to properly entangle `applyDependencySet` aren't all that is needed to solve the issue you described?

test/dependency-manager-adapters/npm-adapter-test.js
```diff
@@ -59,6 +59,43 @@ describe('npmAdapter', () => {
     });
   });
 
+  describe('#changeToDependencySet', () => {
+    it('restores the backup before changing to the new dependency set', () => {
```
rwjblue: I am fairly opposed to this 😸, setting up a new scenario to run should not itself require resetting to the backup first. This resetting seems to be masking another fundamental problem..

huyphamily: Yea, I agree. It does feel a bit strange to reset to the backup first, instead, it should restore the backup at the end of each scenario. Let me look into our particular use case more closely and will get back to you.

dnachev:
@rwjblue The issue is the following:

A project was using `ember-data@3.8.0` (as observed in `yarn.lock`). The tests were passing correct without ember-try, but the scenario, which doesn't touch the `package.json` failed in `ember-try`.

`package.json` has `devDependency`:
`"ember-data": "~3.8.0"`
`yarn.lock` has `ember-data@3.8.0`

And the ember-try scenario was:
3.4 -> `"ember-data": "~3.4.0"`
current -> `no changes`

- The `3.4` scenario runs and replaces `ember-data` in `yarn.lock` with `3.4.1`. It passes.
-  The next scenario is "current", which asks for no changes in `package.json`.  Original `package.json` is restored and `yarn install` is ran. It detects that `yarn.lock` is not valid anymore and re-resolves `ember-data` package again, which ends up picking up `3.8.3`, which is the current `3.8.x` version.
- The test fails, because there is a bug in `3.8.3`, which breaks the application.

The ask is to make sure that `ember-try` doesn't introduce additional variability in the tests in case lockfile is used by `ember-try`, even though these failures might be valid.

rwjblue:
@dnachev - This PR (at the moment) is resetting the various files _before_ starting a scenario, when instead it should be ensuring that the resetting happens when the current scenario is completed.

Specifically, in your timeline:


> * The `3.4` scenario runs and replaces `ember-data` in `yarn.lock` with `3.4.1`. It passes.

Once the scenario passes, the `package.json` and `yarn.lock` should be reset to their original values.

> * The next scenario is "current", which asks for no changes in `package.json`.  Original `package.json` is restored and `yarn install` is ran. It detects that `yarn.lock` is not valid anymore and re-resolves `ember-data` package again, which ends up picking up `3.8.3`, which is the current `3.8.x` version.

The original `package.json` should not restored by running the next ember-try scenario (which is what you are inferring here), it should be restored at the _end_ of the 3.4 scenario.

huyphamily:
@rwjblue yea, I agree. I will refactor the code and move the logic for backup into the try-each.

rwjblue:
I did a bit more digging, and I think the issue is here:

https://github.com/ember-cli/ember-try/blob/bff58592fc44e25e19e534418cf6ef5227763cd8/lib/dependency-manager-adapters/npm.js#L145-L158

As you can see, the original `package.json` contents are always used as the starting point before merging the scenario's specific changes into it. This means that the `package.json` is effectively being reset upon setup of each scenario. However, we do nothing RE: the lockfiles there.

When using `try:one` this discrepancy doesn't really matter too much, because we are only ever running a single scenario which ends with either `--skip-cleanup` or properly restoring the backup files (including lock files), but when using `try:each` those backup files are only restored at the end of _all_ scenarios.

IMHO, the best fix would be to update here:

https://github.com/ember-cli/ember-try/blob/bff58592fc44e25e19e534418cf6ef5227763cd8/lib/tasks/try-each.js#L51-L81

To ensure that we run `this.ScenarioManager.cleanup()` after the command has been run.

We should also deprecate the `--skip-cleanup` option when using `try:each`, I don't think it really provides any value (it is pretty hard to reason about what will be the "final state" when the process exits.

huyphamily:
@rwjblue Yea, I don't have enough context to understand why `--skip-cleanup` was there to begin with, so I rather not touch it. I also don't think we should call cleanup after every scenario in the try each because cleanup involves restoring the backup to the original files (package.json.ember-try -> package.json) and then deleting the backup (package.json.ember-try). Deleting the backup after each scenario is unnecessary, as you only really need to do it at the very end, and adds one more unnecessary step between scenarios because now you have to back up the original dependencies again since you deleted the backup.

I think the easiest thing to do is to call `_restoreOriginalDependencies` when each task finishes inside the `_runCommandForThisScenario` function in try-each.js:
```javascript
return RSVP.resolve(runResults).then(() => task._restoreOriginalDependencies());
```

This is a much easier change and won't affect current behavior. Let me know what you think.

rwjblue:
@huyphamily In general that seems fine (only restoring the `package.json` and lockfiles), but we need to go through the `ScenarioManager` API and not be calling private methods on the underlying adapters directly. You'll need to see if there is an API on `ScenarioManager` to do this, if not add one.

huyphamily:
@rwjblue Code updated. I have added an API to ScenarioManger for restoring the original backup and calling it at the end of each scenario.

rwjblue:
I thought you mentioned that you didn't want to restore the `node_modules/` directory in addition to restoring `package.json`/`yarn.lock`?

huyphamily:
@rwjblue yea, you're right, we don't want to restore the `node_modules/`... Should I separate the backup of every file except the `node_modules` out into another method which will be called in `restoreOriginalDependencies` and keep `restoreOriginalDependencies` as a private method? The new method will not be private.

huyphamily:
@rwjblue or we can pass a parameter into `restoreOriginalDependencies` which determines whether to backup node_modules or not. What are your thoughts?

rwjblue:
Hmm, I thought it we already had a method that restored the manifest and lockfiles but not the fully `node_modules/`. :thinking:

huyphamily:
@rwjblue we do not...unless I miss something?

rwjblue:
I guess I was assuming that the fix here would be to modify `applyDependencySet` to restore the associated lockfile in addition to the `package.json` (we currently restore the original `package.json` because we _start_ with the backup `package.json` every time).

https://github.com/ember-cli/ember-try/blob/bff58592fc44e25e19e534418cf6ef5227763cd8/lib/dependency-manager-adapters/npm.js#L145-L158

rwjblue:
@huyphamily - Friendly ping 😃😃

huyphamily:
@rwjblue Apologies for the super late replies. 

For try:one, we properly restore the backup files, so we can ignore this case.

For, try:each current behavior is to restore the backup files at the end of all scenarios, which is where our issue lies. We want to restore at the end of each scenario, hence why the change to try-each.js.

We do not want to restore any of the associated backup files in `applyDependencySet` because of your earlier statement, which makes complete sense:
```
This PR (at the moment) is resetting the various files before starting a scenario, when instead it should be ensuring that the resetting happens when the current scenario is completed.
```

huyphamily:
@rwjblue any thoughts on this?

rwjblue:
> For try:each, the current behavior is to restore the backup files at the end of all scenarios, which is where our issue lies. We want to restore at the end of each scenario, hence why the change to try-each.js.

The thing is, I don't really want to restore the `node_modules` folder **and** the various individual files (e.g. `yarn.lock`, shrinkwrap, etc). However, I think that particular thing could be something we change in the future (I'll make an issue for it) and doesn't need to be coupled to this PR.

huyphamily:
> There are still a few changes needed here to ensure that all of the built in managers work properly with the main change.
> 
> * Update the `bower` adapter to implement `restoreOriginalDependencies` (would need to be extracted from current `cleanup` logic):
> 
> https://github.com/ember-cli/ember-try/blob/aa4e3fff85b538f3b1aa77b00af0add6bc7d6f94/lib/dependency-manager-adapters/bower.js#L93-L110
> 
> * Update `workspace` manager to implement `restoreOriginalDependencies`, this will look _basically_ the same as the `cleanup` method (but for `restoreOriginalDependencies`):
> 
> https://github.com/ember-cli/ember-try/blob/aa4e3fff85b538f3b1aa77b00af0add6bc7d6f94/lib/dependency-manager-adapters/workspace.js#L88-L90

I just got back from paternity leave and things got a little hectic, so sorry for the late reply. 

I made the changes to add restoreOriginalDependencies in workplace adapter. However, I did not add restoreOriginalDependencies for bower since you mention in the past that not to add to bower as it's going away. Is this not still true?

rwjblue:
> However, I did not add restoreOriginalDependencies for bower since you mention in the past that not to add to bower as it's going away. Is this not still true?

Ya, it is true in general, but without doing _something_ here we are going to throw an error if you have Bower scenarios and _that_ is not ok...

rwjblue:
I've rebased and made a tweak to ensure that when running `ember try:one <scenario> --skip-cleanup` we don't actually cleanup.

I'll be working on finalizing this PR today, I think at this point we just need lots of tests:

* using multiple scenarios, confirm that the package.json's are not gradually mutated after each (reported here and in #579)
* using a single scenario with `--skip-cleanup`, confirm that cleanup is not actually done

bertdeblock:
Closing for now.

PR #467, opened 2020-03-12
Ensure Yarn Workspace adapter supports `ember try:reset`.

### LINKED ISSUE #466: `ember try:reset` with workspaces
I've tried running `ember try:reset` within a monorepo with yarn workspaces and I'm seeing the following error

```
TypeError: Cannot read property 'map' of undefined
/node_modules/ember-try/lib/dependency-manager-adapters/workspace.js:90:43
/node_modules/ember-try/lib/utils/scenario-manager.js:40:25
...
```

It looks like `this._packageAdapters` is `undefined`?

https://github.com/ember-cli/ember-try/blob/aa4e3fff85b538f3b1aa77b00af0add6bc7d6f94/lib/dependency-manager-adapters/workspace.js#L88-L90

It looks like the package adapters are initialised in `setup()`

https://github.com/ember-cli/ember-try/blob/aa4e3fff85b538f3b1aa77b00af0add6bc7d6f94/lib/dependency-manager-adapters/workspace.js#L52-L60

In the tests `setup()` is run before `cleanup()`, so either the tests are assuming `setup()` is run first and it isn't, or the package adapters should be initialised in both `setup()` and `cleanup()`?

https://github.com/ember-cli/ember-try/blob/aa4e3fff85b538f3b1aa77b00af0add6bc7d6f94/test/dependency-manager-adapters/workspace-adapter-test.js#L210-L219

I can put together a PR, but I'm still figuring out what order things run in..

rwjblue:
Fundamentally, `Adapter.prototype.cleanup` **cannot** require `Adatper.prototype.setup` to have been called. _Most_ of the time `.setup` is called before `.cleanup` (this happens with `ember try:each` and `ember try:one`), but when `--skip-cleanup` is specified (e.g.  `ember try:one scenario-name --skip-cleanup`) and you subsequently run `ember try:reset` the `.setup` happened in a completely different process.

The issue here is a misunderstanding when we added the `WorkspaceAdapter` initially. Basically, the assumtion was that `.setup` was for setting up both local instance state (e.g. the `this._packageAdapters` array) **and** for actually preparing for a `ember try:...` run (doing backups / etc).

The fix is to move the internal state setup code directly into `init`/`constructor, and ensure that `.setup` is only used to create the backups themselves.

Fixes #466

### DIFF
--- a/lib/dependency-manager-adapters/workspace.js
+++ b/lib/dependency-manager-adapters/workspace.js
@@ -18,15 +18,6 @@ module.exports = CoreObject.extend({
     if (!this.useYarnCommand) {
       throw new Error('workspaces are currently only supported by Yarn, you must set `useYarn` to true');
     }
-  },
-
-  packageJSON: 'package.json',
-
-  setup(options) {
-    if (!options) {
-      options = {};
-    }
-
     let packageJSON = JSON.parse(fs.readFileSync(this.packageJSON));
     let workspaceGlobs;
 
@@ -58,6 +49,15 @@ module.exports = CoreObject.extend({
         buildManagerOptions: this.buildManagerOptions,
       });
     });
+
+  },
+
+  packageJSON: 'package.json',
+
+  setup(options) {
+    if (!options) {
+      options = {};
+    }
 
     return RSVP.all(this._packageAdapters.map(adapter => adapter.setup(options)));
   },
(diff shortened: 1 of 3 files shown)


BarryThePenguin:
Works well now, thank you 👏

timiyay:
@rwjblue Any chance of a new release to include the changes in this PR?

We're keen to upgrade from 3.12 LTS to 3.16 LTS, and we'd like to have `ember-try` in place to help us find issues, but are currently blocked by this yarn workspaces issue.

PR #512, opened 2020-08-11
Drop support for Node.js 6.x, 8.x, 11.x and 13.x

Turbo87:
which should match our support matrix to all of the currently officially supported Node.js versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
   npm-smoke-tests:
     strategy:
       matrix:
-        node: ['8', '10', '12', '14']
+        node: ['10', '12', '14']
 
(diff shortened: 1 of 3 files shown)


Turbo87:
@kategengler should be fixed now, but it seems that CircleCI is now failing for unrelated reasons 😞 

any suggestions?

kategengler:
I just hit "rerun" -- a little concerning that it only fails on the modern Embers. It's possible we should update the scenarios so that it doesn't run so many old ones, but that is unrelated to the failures.

kategengler:
🤔 all-commands-smoke-test uses npm (vs yarn)

Turbo87:
@rwjblue @kategengler looks like this is ready to go now :)

kategengler (approved):

PR #583, opened 2020-10-28
Update to use async/await modern language features.

rwjblue:


package.json
```diff
@@ -42,12 +42,10 @@
     "debug": "^4.1.1",
     "ember-try-config": "^3.0.0",
     "execa": "^1.0.0",
-    "extend": "^3.0.0",
     "fs-extra": "^5.0.0",
     "promise-map-series": "^0.3.0",
```
kategengler, line 46: Think the PR removes the last use of `promise-map-series`

kategengler (approved):

package.json
rwjblue: Ya, I think so too, I missed that one when I was removing the others.

PR #584, opened 2020-10-28
Use matrix builds to speed up all commands smoke tests

rwjblue:
Instead of running a single shell script, run a job for each of the individual commands. This will result in many more CI jobs, but _should_ be much much faster total time.


package.json
```diff
@@ -24,16 +24,15 @@
...
     "build": "ember build",
-    "client-test": "./run-smoke-test.sh all-commands.sh",
     "lint": "eslint lib test",
     "lint:hbs": "ember-template-lint .",
     "lint:js": "eslint .",
     "node-test": "mocha test/**/*.js",
     "node-test-with-coverage": "nyc --reporter lcov npm run-script node-test && node_modules/.bin/codecov",
-    "smoke-test": "./run-smoke-test.sh smoke-test.sh",
-    "smoke-test-yarn": "./run-smoke-test.sh smoke-test-yarn.sh",
+    "smoke-test": "cd smoke-test-app && ./smoke-test.sh",
```
kategengler, line 32: I think these scripts are "weird" now locally because unless you've gone into `smoke-test-app` and installed with the correct package manager, you can easily run `smoke-test-yarn` and have it being using npm, and vice versa.

kategengler (approved):

package.json
rwjblue: I just decided to delete these scripts all together. I couldn't think of a way to ensure they actually run in the right contexts, and since CI is actually pretty fast / peppy it's _probably_ fine...

PR #593, opened 2020-10-29
Migrate Windows smoke test to GitHub Actions.

rwjblue:


rwjblue:
AppVeyor CI build will fail, but once this is merged I'll delete the AppVeyor webook and then future PR's won't fail like that.

PR #745, opened 2021-10-01
CI: Remove job schedule

Turbo87:
Scheduled jobs are automatically disabled after a certain amount of days, thanks to the cryptocurrency community.

This commit removes the schedule to keep the job enabled for PRs and regular commits

/cc @kategengler


kategengler:
Thank you!

PR #788, opened 2021-11-04
tests/npm-adapter: Use async/await instead of promise chains

Turbo87:
This should make the test code a little easier on the eyes :)


kategengler:
Thank you!

PR #821, opened 2022-02-03
Use `dereference` when copying node_modules

### LINKED ISSUE #813: Error: Cannot copy '../semver/bin/semver.js' to a subdirectory of itself
If you run `test:ember-compatibility` more than once, it fails with the following:

```
Error!
Error: Cannot copy '../semver/bin/semver.js' to a subdirectory of itself, '../semver/bin/semver.js'.
Error: Cannot copy '../semver/bin/semver.js' to a subdirectory of itself, '../semver/bin/semver.js'.
    at /home/*****/*****/e4a/node_modules/fs-extra/lib/copy/copy.js:210:21
    at FSReqCallback.oncomplete (node:fs:188:23)
```

Steps to reproduce:

```
ember addon e4a
cd e4a
npm install
npm run test:ember-compatibility
num run test:ember-compatibility
```

I had a co-worker try the same steps and he got the same error.

kategengler:
- `clobber` option to fs-extra#copy was renamed to `overwrite` way back in v2
- Upgrade fs-extra

Fixes #813

This can also be fixed by removing the backing up and restoring of node_modules, which we suspect is no longer necessary. The only use case I recalled for it was restoring node_modules exactly (including any linked packages), which this change would already no longer support.

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -217,7 +217,10 @@ module.exports = CoreObject.extend({
     let nodeModulesBackupLocation = path.join(this.cwd, this.nodeModulesBackupLocation);
     if (fs.existsSync(nodeModulesBackupLocation)) {
       restoreTasks.push(
-        copy(nodeModulesBackupLocation, path.join(this.cwd, this.nodeModules), { clobber: true })
+        copy(nodeModulesBackupLocation, path.join(this.cwd, this.nodeModules), {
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

kategengler:
Based on the chat yesterday, I'm going to PR removing the backup of node_modules in place of this PR

bertdeblock:
@kategengler Could we maybe land these changes in the meanwhile?

PR #843, opened 2022-04-14
Fix copy options

### LINKED ISSUE #813: Error: Cannot copy '../semver/bin/semver.js' to a subdirectory of itself

wagenet:
There is no `clobber` option. `overwrite` appears to do what was intended.

I had hoped this would be a fix for #813, but I'm still having issues :/

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -217,7 +217,9 @@ module.exports = CoreObject.extend({
     let nodeModulesBackupLocation = path.join(this.cwd, this.nodeModulesBackupLocation);
     if (fs.existsSync(nodeModulesBackupLocation)) {
       restoreTasks.push(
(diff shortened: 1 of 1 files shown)


bertdeblock:
Think this can be closed due to https://github.com/ember-cli/ember-try/pull/962 being merged.

PR #850, opened 2022-05-13
Add support for npm overrides

mansona:
This PR is supposed to help with the issue that is described in https://github.com/emberjs/ember-render-modifiers/issues/64

Essentially npm thinks that ranges shouldn't be satisfied if you are using pre-releases. You can overcome this if you use an `overrides` in npm which works in a very similar way to yarn resolutions. 

The issue is that if you provide an `overrides` config to ember-try it will just ignore it: 

```
{
  name: 'ember-release',
  npm: {
    devDependencies: {
      'ember-source': await getChannelURL('release'),
    },
    overrides: {
      'ember-source': '$ember-source',
    },
  },
},
```

This PR implements npm overrides in much the same way as yarn resolutions 👍  in fact I copy and pasted all of the resolutions tests and just changed resolutions to overrides in most cases 😂

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,14 @@ module.exports = async function() {
             'ember-source': '2.11.0'
           },
           /*
+            You can optionally define npm overrides to enforce a  specific dependency version
+            to be installed. This is useful if other libraries you depend on include different
+            versions of a package. This does nothing if `useYarn` is true;
+          */
+          overrides: {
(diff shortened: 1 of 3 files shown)


kategengler:
Thanks! Can you also add something to the readme about this feature?

kategengler:
Thank you!

PR #896, opened 2022-11-09
Fixed version of ember-try-test-suite-helper

candunaj:
on newer version of npm it was throwing an error because of incorrect version in a fixture.

This PR fixes it


mansona:
We actually fixed this by just deleting that dependency 🤷 it wasn't being used or checked anywhere 🙃

PR #912, opened 2023-02-07
Downgraded fs-extra version to 6.x to fix a fs.copy related issue.

larry-x-yu:
fs-extra brought in a breaking change here:
https://github.com/jprichardson/node-fs-extra/pull/582

where it breaks ember-try when copying node_modules/ and there's a symlink in it. The bug would either prevent ember-try from running or cleaning up after try-each.

Testing done:
`yarn test` passed locally.

@Co-authored-by hjdivad


hjdivad:
@larry-x-yu i rebased your branch after merging #913

PR #913, opened 2023-02-08
Bump node to supported versions

hjdivad:
See [Node Releases](https://github.com/nodejs/release#release-schedule)

Floating deps test is currently failing with things like

> The engine "node" is incompatible with this module. Expected version "^12.13.0 || ^14.15.0 || >=16.0.0". Got "10.24.1"

Which is to be expected for out-of-support node versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
+        with:
+          node-version: 14
 
       - run: yarn install
       - run: yarn lint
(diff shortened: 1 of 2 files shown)


rwjblue (approved):

SergeAstapov:
@hjdivad  @rwjblue should we bumo even to 16+? Node.js 14 is EOL in less then two months from now and going 16+ will prevent the same issue in future for more time

hjdivad:
@SergeAstapov This makes sense to me.  This PR is only cut in a pre-release, so having 3.0.0 proper only support node 16 sounds right.

Do you have cycles to post a PR? I'd be happy to review if you do.

SergeAstapov:
@hjdivad done! see #920

PR #920, opened 2023-03-02
Drop Node.js 14 support

SergeAstapov:
Follow up to #913

### DIFF
--- a/lib/dependency-manager-adapters/npm.js
+++ b/lib/dependency-manager-adapters/npm.js
@@ -8,6 +8,7 @@ const path = require('path');
 const debug = require('debug')('ember-try:dependency-manager-adapter:npm');
 const rimraf = util.promisify(require('rimraf'));
 const chalk = require('chalk');
+const semver = require('semver');
 
 module.exports = CoreObject.extend({
   init() {
@@ -202,6 +203,15 @@ module.exports = CoreObject.extend({
         delete packageJSON[kindOfDependency][packageName];
       } else {
         packageJSON[kindOfDependency][packageName] = version;
+
+        // in npm we need to always add an override if the version is a pre-release
+        if (semver.prerelease(version) || /^https*:\/\/.*\.tg*z/.test(version)) {
+          if (!packageJSON.overrides) {
+            packageJSON.overrides = {};
+          }
+
+          packageJSON.overrides[packageName] = `$${packageName}`;
+        }
       }
     });
   },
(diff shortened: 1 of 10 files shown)


hjdivad (commented):
Can you look into [the failing tests](https://github.com/ember-cli/ember-try/pull/920/checks#step:5:121)

As in #913 the floating deps test is expected to fail, but the rest should be green.

hjdivad:
@SergeAstapov

SergeAstapov:
@hjdivad it seems never version of npm bundled with Node.js 16 cause a hard error because `ember-try-test-suite-helper@0.0.1` [specific here](https://github.com/ember-cli/ember-try/blob/master/test/fixtures/package.json) actually does not exist on npm (only 1.0.0 and 1.0.1 have been published)

it seems more correctly to actually publish that version to npm, wonder if @kategengler can do so.

meanwhile I'll try to look if there is something else missing or I can came up with a workaround

mansona:
Hey folks 👋  https://github.com/ember-cli/ember-try/pull/893 has CI passing and works around that problem you also identified @SergeAstapov

lib/dependency-manager-adapters/npm.js
```diff
@@ -202,6 +203,15 @@ module.exports = CoreObject.extend({
         delete packageJSON[kindOfDependency][packageName];
       } else {
         packageJSON[kindOfDependency][packageName] = version;
+
+        // in npm we need to always add an override if the version is a pre-release
```
kategengler, line 207: There isn't a separate adapter for yarn, so this will add it for that as well

lib/dependency-manager-adapters/npm.js
```diff
@@ -202,6 +203,15 @@ module.exports = CoreObject.extend({
...
+        if (semver.prerelease(version) || /^https*:\/\/.*\.tg*z/.test(version)) {
+          if (!packageJSON.overrides) {
+            packageJSON.overrides = {};
+          }
+
+          packageJSON.overrides[packageName] = `$${packageName}`;
```
kategengler, line 213: We should add tests for this behavior

kategengler (changes requested):
Changes requested from @mansona

lib/dependency-manager-adapters/npm.js
mansona: done!

smoke-test-app/tests/dummy/config/ember-try.js
```diff
@@ -28,6 +30,9 @@ module.exports = async function () {
           devDependencies: {
             'ember-source': await getChannelURL('release'),
           },
+          overrides: {
```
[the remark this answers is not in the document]
mansona: yes good catch! I had forgotten they were there 🙈

kategengler (approved):

kategengler:
Thank you!!

PR #934, opened 2023-03-22
Update github CI yaml

mansona:
This just takes some inspiration from the default output of a new Ember addon 👍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,14 @@ name: CI
 on:
   push:
     branches:
+      - main
       - master
       - 'v*'
   pull_request: {}
+
+concurrency:
+  group: ci-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,32 +25,38 @@ jobs:
 
       - run: yarn install
       - run: yarn lint
-
-  tests:
-    name: Tests
+  test:
+    name: "Tests"
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
kategengler, line 31: I suspect these timeouts won't work for the test suite

mansona: So this timeout is only for that particular test, and that test seems to be running in less than a minute so 10 mins is not so bad https://github.com/ember-cli/ember-try/actions/runs/4495038220/jobs/7908231867

The issue that we have is that any tests that hang will time out after **6 hours!!** 😱 https://github.com/ember-cli/ember-try/actions/runs/4486271160/jobs/7888744453#step:6:1 

I don't know who decided 6 hours was a reasonable default for a timeout but this PR now uses the "ember addon default" of 10 mins for normal tests and 20 mins for the longer smoke tests. And that will allow each test to run for 20 mins not a total running time for all scenarios 👍

mansona: The last run of this is a great example of what I mean. Because I fixed CI in https://github.com/ember-cli/ember-try/pull/933 all the dependabot PRs updated at the same time and drained the available workers for this repo to run CI. The elapsed time of the last run was quite high because it was waiting for workers but only each individual job was bound by its own timeout 👍

PR #957, opened 2023-06-24
Document pnpm options and behavior

NullVoxPopuli:
RFC: https://github.com/emberjs/rfcs/pull/907
Implementation PR: https://github.com/ember-cli/ember-cli/pull/10287
Advancement PR: https://github.com/emberjs/rfcs/pull/932

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,11 @@ module.exports = async function() {
       dependencies will be restored to their prior state.
     */
     useYarn: true,
+    /*
+      If set to true, all npm scenarios will use `pnpm` for install with the `--frozen-lockfile=false` option. At cleanup, your
+      dependencies will be restored to their prior state.
+    */
+    usePnpm: true,
 
     /*
       buildManagerOptions allows you to opt-out of the default options such as `--ignore-engines --no-lockfile`.
@@ -144,7 +149,7 @@ module.exports = async function() {
             'ember-source': '2.11.0'
           },
           /*
-            You can optionally define npm overrides to enforce a  specific dependency version
+            You can optionally define npm or pnpm overrides to enforce a  specific dependency version
             to be installed. This is useful if other libraries you depend on include different
             versions of a package. This does nothing if `useYarn` is true;
           */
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,6 +120,11 @@ module.exports = async function() {
       dependencies will be restored to their prior state.
     */
     useYarn: true,
+    /*
+      If set to true, all npm scenarios will use `pnpm` for install with the `--frozen-lockfile=false` option. At cleanup, your
```
kategengler, line 124: Is `--frozen-lockfile` the same as ignoring the lockfile?

kategengler:
Thanks!

README.md
NullVoxPopuli: not quite, but it'll allow things to float a bit and update the lockfile if package.json's change.
https://pnpm.io/cli/install#--frozen-lockfile

There isn't exactly a `--no-lockfile` option, like what yarn has, though there is a `--fix-lockfile` option: https://pnpm.io/cli/install#--fix-lockfile (docs are very light there)

I didn't have any problems with the currently implemented behavior, but something goofy I've noticed with pnpm in general is that when it resolves `^` or `~` ranges, it resolves first to what your have in your "store" (for speed) -- this can be somewhat annoying when I want "latest" -- but also, I heavily rely on automated dependency updating for meaningful dep updates.

kategengler: That may somewhat defeat the point of ember-try; I wonder if ember-try should by default put the specific version in `overrides` or `resolutions`.

NullVoxPopuli: yeah, I think for consistent, it'd be cool to [`await latestVersion(range)`](https://www.npmjs.com/package/latest-version) on the try-config versions -- because at least then we can assure consistent behavior regardless of package manager. (anyone tried if yarn3/4 does anything goofy? :sweat_smile: )

kategengler: You could do that today in a try config, but if it isn't set in resolutions or overrides, you'll get whatever the package manager gives you for the version that is set.

NullVoxPopuli: yeep

PR #958, opened 2023-06-25
Update and test pnpm behavior so that we simulate a new user install

NullVoxPopuli:
Based on discussion here: https://discord.com/channels/480462759797063690/480501885837770763/1122582845873983529

### DIFF
--- a/test/dependency-manager-adapters/pnpm-adapter-test.js
+++ b/test/dependency-manager-adapters/pnpm-adapter-test.js
@@ -34,6 +34,7 @@ describe('pnpm Adapter', () => {
       expect(await fs.readJson('package.json.ember-try')).to.deep.equal({
         originalPackageJSON: true,
       });
+      expect(await fs.pathExists('pnpm-lock.yaml')).to.be.false;
       expect(await fs.readFile('pnpm-lock.ember-try.yaml', 'utf-8')).to.equal(
         'originalYAML: true\n'
       );
--- a/test/dependency-manager-adapters/pnpm-adapter-test.js
+++ b/test/dependency-manager-adapters/pnpm-adapter-test.js
@@ -41,6 +41,16 @@ describe('pnpm Adapter', () => {
       expect(fs.existsSync('.node_modules.ember-try')).to.be.false;
     });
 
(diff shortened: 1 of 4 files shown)


kategengler:
I wonder if we should make this new behavior have an opt-in or out? 

Users have the capability to override `--ignore-lockfile` for the other managers if they do want to test _with_ a lockfile.

NullVoxPopuli:
that's a good point :thinking: 

are you thinking it'd be like, a top-level config option like `usePnpm`?

like, maybe:
```js
{
  usePnpm: true,
  adapterOptions: { // specific to each adapter, with their own defaults
    keepLockfile: true, // default false
  }

}
```
idk

kategengler:
@NullVoxPopuli I think that's a bit too close to `buildManagerOptions`. I think it should be opt-out, same as the other managers (though theirs are through flags). I don't really love adding more top-level config, but maybe we should here: `floatDependencies: true`. We could make it work for the other managers too, via their command line flags, though it would interact unfortunately with `buildManagerOptions` and `npmOptions`. 

Alternatively, I think there's some precedent in the repo for making our own flags that can be supplied in `npmOptions` or `buildManagerOptions` that we remove before passing them along.

test/dependency-manager-adapters/pnpm-adapter-test.js
```diff
@@ -34,12 +34,23 @@ describe('pnpm Adapter', () => {
...
         originalPackageJSON: true,
       });
+      expect(await fs.pathExists('pnpm-lock.yaml')).to.be.false;
       expect(await fs.readFile('pnpm-lock.ember-try.yaml', 'utf-8')).to.equal(
         'originalYAML: true\n'
       );
       expect(fs.existsSync('.node_modules.ember-try')).to.be.false;
     });
 
+    it('copies up the `pnpm-lock.yaml` file when pnpmUseLockfile is true (non-default)', async () => {
```
NullVoxPopuli, line 44: the inverse of this test is covered by the test above

NullVoxPopuli:
Going to do a separate PR that uses this: 
image

PR #959, opened 2023-06-28
Fix CI

mansona:


mansona:
I was trying to extract this from #920 but actually it's a kinda pointless PR 🙈 we're starting to get npm peer dependency issues in that other PR because we have upgraded to a version that actually checks it. I am pushing the required changes to that PR directly 👍

PR #966, opened 2023-09-14
ember-try: Enforce resolution-mode=highest for pnpm versions 8.0.0 to 8.6.*

lolmaus:
pnpm versions 8.0.0 to 8.6.* had the `resolution-mode` setting inverted which caused troubles with unexpected versions of dependencies pulled for `ember-try` tests.

This is an attempt to fix those issues by enfrocing `resolution-mode=highest` via project-local `.npmrc`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,9 @@ If you include `useYarn: true` in your `ember-try` config, all npm scenarios wil
 
 If you include `usePnpm: true` in your `ember-try` config, all npm scenarios will use `pnpm` for install with the `--no-lockfile` options. At cleanup, your dependencies will be restored to their prior state.
 
+> ⚠ pnpm versions from 8.0.0 to 8.6.x have the default value of [resolution-mode](https://pnpm.io/npmrc#resolution-mode) setting changed to `lowest-direct`. This violates `ember-try` expectations as `resolution-mode` is expected to be `highest`, like in `npm` and `pnpm` versions < 8.0.0 and >= 8.7.0.
+>
+> If you run into this issue, we recommend to upgrade pnpm to latest version. If you are unable to upgrade, you can set `resolution-mode = highest` in the `.npmrc` file.
 
 ##### A note on npm scenarios with lockfiles
 
--- a/lib/dependency-manager-adapters/pnpm.js
+++ b/lib/dependency-manager-adapters/pnpm.js
@@ -5,14 +5,11 @@ const fs = require('fs-extra');
 const path = require('path');
 const debug = require('debug')('ember-try:dependency-manager-adapter:pnpm');
 const Backup = require('../utils/backup');
-const { promisify } = require('util');
-const exec = promisify(require('child_process').exec);
 const semverLt = require('semver/functions/lt');
 const semverGte = require('semver/functions/gte');
 
 const PACKAGE_JSON = 'package.json';
 const PNPM_LOCKFILE = 'pnpm-lock.yaml';
-const NPMRC_CONFIG = '.npmrc';
 
 module.exports = CoreObject.extend({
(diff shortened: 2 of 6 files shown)


lolmaus:
@bertdeblock Mind having a look? 

I did my best to use the Backup API correctly. But I needed to make a couple of small updates into the Backup util.

lolmaus:
CC @mansona

kategengler:
I understand the motivation but this seems a bit overkill. What if we just warn or error on the affected pnpm versions?

cc @NullVoxPopuli

NullVoxPopuli:
I would prefer warn/error as well.

Folks using `pnpm` should (hopefully) already be used to frequent updates and be striving for being on the latest version

NullVoxPopuli:
idea?:

when in the "danger range", check:
```js
$`npm config get resolution-mode` === 'highest'
```
and if that is false, throw the error/warn

bertdeblock:
Personally, I would do nothing. I don't feel it's ember-try's responsibility to warn about, or error on this.

mansona:
> Personally, I would do nothing. I don't feel it's ember-try's responsibility to warn about, or error on this.

I strongly disagree on this one. If we do nothing in this case then the default `embroiderSafe()` and `embroiderOptimised()` ember-try scenarios on addons will be broken without people knowing that they are broken. 

Just to be explicit (I'm not trying to tell you anything you already know) but `embroiderSafe()` from `@embroider/test-setup` currently adds a `^3.0.1` version dependency on `@embroider/core` `@embroider/compat` and `@embroider/webpack` and that means that with `resolution-mode=lowest-direct` you will not be testing your addons against the latest Embroider. This is just one example where the expected behaviour is just wrong and people need to be informed if ember-try is doing something that isn't what is expected.

bertdeblock:
I understand what it means to use `lowest-direct`. I'm just failing to see why ember-try should error on this. Seems like the user's responsibility, as they choose to use pnpm v8. What if the user wants to use `lowest-direct` (either implicit or explicitly defined)?

lolmaus:
> What if the user wants to use lowest-direct (either implicit or explicitly defined)?

Then they may receive incorrect test results with `ember-try`?

For `ember-try` tests to be consistent, the `resolution-mode` must be identical for an addon developer, addon consumer and CI.

bertdeblock:
Maybe, but `lowest-direct` is a valid resolution mode and if a user chooses to (explicitly) use that mode, I feel ember-try should respect that.

lolmaus:
@bertdeblock We're gonna have an Ember Tooling Team meeting today at 17:00 CEST at [Ember Discord](https://discord.com/invite/emberjs). I'm gonna bring this matter up again per your objection.

Please join and contribute to the discussion. 🙏 I'm `lolmaus` at Discord.

kategengler:
> > Personally, I would do nothing. I don't feel it's ember-try's responsibility to warn about, or error on this.
> 
> I strongly disagree on this one. If we do nothing in this case then the default `embroiderSafe()` and `embroiderOptimised()` ember-try scenarios on addons will be broken without people knowing that they are broken.
> 

I think we could get pretty far by noting the version issue / resolution configuration in the README and publicizing it. For a fix to be protective going forward users would need to update ember-try so this won't fix anything for those leaving their addons alone for awhile. Support for `pnpm` is relatively new and those using it before official support landed in ember-cli (5.3) are more likely to be the type of user that is aware of this problem. 

@bertdeblock's point that users may choose that type of resolution is a very good one. Our README should warn that if that mode is chosen, they may be losing the benefits of ember-try unless they explicitly have scenarios for each version.

mansona (approved):
Looks good to me, and it matches what we discussed in the meeting yesterday (I think) 🎉

PR #999, opened 2024-12-10
Remove use of `core-object`

bertdeblock:


lib/commands/reset.js
```diff
@@ -10,7 +10,6 @@ module.exports = {
     let ResetTask = require('../tasks/reset');
 
     let resetTask = new ResetTask({
-      ui: this.ui,
```
bertdeblock, line 13: Not used in `ResetTask`.

kategengler (approved):

PR #1023, opened 2024-12-11
Use `ember version` command to make test output more readable

bertdeblock:


test/tasks/try-each-test.js
```diff
@@ -632,7 +632,7 @@ describe('tryEach', () => {
           ui: { writeLine: outputFn },
           project: { root: tmpdir },
           config,
-          commandArgs: ['ember', 'help', '--json', 'true'],
```
bertdeblock, line 635: `ember help --json` outputs a very long JSON string, which makes the output harder to read when testing locally.

kategengler (approved):

Issue #970, opened 2024-10-16
Overrides do no work in PNPM monorepo


jelhan:
PNPM supports `pnpm.overrides` and `resolutions` only at the root of a project. Ember Try sets it at the closest `package.json`. In a typical setup of a v2 addon that's the "test-app" package. That does not have any affect. Ember Try must set it in the `package.json` at the workspace root.

Likely there is a second bug: Ember Try sets `overrides` at the root of the package.json. That's correct for NPM. But I think for PNPM it must go into a `pnpm` member. This is an example taken from PNPM documentation:

```json
{
  "dependencies": {
    "foo": "^1.0.0"
  },
  "pnpm": {
    "overrides": {
      "foo": "$foo"
    }
  }
}
```

But Ember Try generates the following, which is valid for NPM but does not fit with PNPM documentation:

```json
{
  "dependencies": {
    "foo": "^1.0.0"
  },
  "overrides": {
    "foo": "$foo",
  }
}
```

kategengler:
Using pnpm in a monorepo I see: 
`WARN  The field "pnpm.overrides" was found in test-app/package.json. This will not take effect. You should configure "pnpm.overrides" at the root of the workspace instead.`

So I think pnpm is telling us exactly what it wants.

kategengler:
I'm not sure setting overrides correctly at the root solves the peerDependency errors seen when using embroider + pnpm + pre-release versions.

PR #1026, opened 2024-12-12
Use a completely separate adapter for yarn

bertdeblock:
Builds on #1025.

This separates the yarn-specific logic from the npm-specific logic.
This also introduces a `BaseAdapter` class to get rid of duplication and ensure more consistency among adapters.


lib/dependency-manager-adapters/base.js
```diff
@@ -0,0 +1,188 @@
...
+    await this.backup.addFiles([PACKAGE_JSON, this.lockfile]);
+  }
+
+  async changeToDependencySet(dependencySet) {
+    await this.applyDependencySet(dependencySet);
+    await this._install(dependencySet);
+
+    const dependencies = {
+      ...dependencySet.dependencies,
+      ...dependencySet.devDependencies,
```
kategengler, line 49: For a future addition, I wonder if this should also do peerDependencies

lib/dependency-manager-adapters/base.js
```diff
@@ -0,0 +1,188 @@
...
+    const oldPackageJSON = JSON.parse(fs.readFileSync(this.backup.pathForFile(PACKAGE_JSON)));
+    const newPackageJSON = this._packageJSONForDependencySet(oldPackageJSON, dependencySet);
+
+    this.debug('Write package.json with: \n', JSON.stringify(newPackageJSON));
+
+    fs.writeFileSync(path.join(this.cwd, PACKAGE_JSON), JSON.stringify(newPackageJSON, null, 2));
+
+    // We restore the original lockfile here, so that we always create a minimal
+    // diff compared to the original locked dependency set:
+    await this.backup.restoreFile(this.lockfile);
```
kategengler, line 80: I think this is a significant change in behavior? Restoring the lockfile before the install is no different than leaving it in place and editing the package.json. 

It is debated often, but we've had everything 'float' in ember-try runs, both so that packages test against the newest versions of their in-range dependencies but also to help with situations where other deps need to float to enable the changed packages in the scenario.

lib/dependency-manager-adapters/pnpm.js
```diff
@@ -1,70 +1,18 @@
...
+const { LOCKFILE } = require('../utils/package-managers');
+const { BaseAdapter } = require('./base');
 
-const PACKAGE_JSON = 'package.json';
-const PNPM_LOCKFILE = 'pnpm-lock.yaml';
+module.exports = class PnpmAdapter extends BaseAdapter {
+  defaultInstallOptions = ['--no-lockfile', '--ignore-scripts'];
+  lockfile = LOCKFILE.pnpm;
+  name = 'pnpm';
+  overridesKey = 'overrides';
```
kategengler, line 11: pnpm overrides key is [nested in a `pnpm` key](https://pnpm.io/package_json#pnpmoverrides) and afaict they do not support [top-level overrides](https://github.com/pnpm/pnpm/issues/7421). 

This bug was here before https://github.com/ember-cli/ember-try/issues/970 and I wonder if it isn't the source of the issue where sometimes people see issues with peer dependencies.

bertdeblock: The pnpm adapter had this, so I thought it would make sense for all adapters. Will remove!

lib/dependency-manager-adapters/base.js
bertdeblock: The pnpm adapter had this, so I thought it would make sense for all adapters. Will remove!

lib/dependency-manager-adapters/pnpm.js
bertdeblock: Will fix!

kategengler: thank you <3 

I know it is slightly tangential but your nice refactor incidentally makes it a touch harder to fix, so it seemed to fit now

lib/dependency-manager-adapters/base.js
kategengler: ah weird, that must have snuck by

.eslintrc.js
```diff
@@ -39,10 +39,10 @@ module.exports = {
         browser: false,
         node: true
       },
-      extends: ['plugin:node/recommended'],
```
bertdeblock, line 42: eslint-plugin-node was causing false errors.

lib/dependency-manager-adapters/pnpm.js
bertdeblock: Should be fixed now.

package.json
```diff
@@ -42,6 +43,7 @@
     "walk-sync": "^2.2.0"
   },
   "devDependencies": {
+    "@babel/core": "^7.26.0",
```
bertdeblock, line 46: Required peer dep of `@babel/eslint-parser`.
Forgot to add this when updating the ESLint setup.

kategengler (approved):

PR #1029, opened 2024-12-17
Remove use of `mockery`

bertdeblock:
The `mockery` project is old, and archived.
I also had some issues with it when converting the test suite to ES modules, so seemed easiest to just get rid of it, and use a more "vanilla" approach.

### DIFF
--- a/lib/utils/run.js
+++ b/lib/utils/run.js
@@ -3,7 +3,9 @@
 const execa = require('execa');
 const debug = require('debug')('ember-try:utils:run');
 
-module.exports = async function run(command, args, _options) {
+let runFunction = originalRunFunction;
+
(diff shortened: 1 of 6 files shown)


lib/utils/run.js
```diff
@@ -21,4 +23,22 @@ module.exports = async function run(command, args, _options) {
...
+  runFunction = mockedRunFunction;
+}
+
+function restoreRun() {
+  runFunction = originalRunFunction;
+}
+
+module.exports = run;
+
+module.exports._originalRunFunction = originalRunFunction;
```
bertdeblock, line 42: Not ideal though, to have these "private" exports.

kategengler (approved):

PR #1034, opened 2024-12-18
Migrate test suite to ES modules

bertdeblock:


.eslintrc.js
```diff
@@ -4,6 +4,10 @@ module.exports = {
     {
       env: { mocha: true },
       files: ['test/**/*.js'],
+      parserOptions: { sourceType: 'module' },
+      rules: {
+        'n/no-extraneous-import': 'off',
```
bertdeblock, line 9: I assume because we use a "fake" `package.json` file to force ES modules for the `test` folder. Can probably be deleted once we use `"type": "module"` in the root `package.json` file.

package.json
```diff
@@ -51,17 +51,17 @@
     "@babel/core": "^7.26.0",
     "@babel/eslint-parser": "^7.25.9",
     "@release-it-plugins/lerna-changelog": "^7.0.0",
-    "chai": "^4.4.1",
-    "chai-as-promised": "^7.1.1",
+    "chai": "^5.1.2",
```
bertdeblock, line 54: We can now use the latest versions of our test dependencies.

test/dependency-manager-adapters/workspace-adapter-test.js
```diff
@@ -1,14 +1,13 @@
...
+import { expect } from 'chai';
+import fs from 'fs-extra';
+import path from 'path';
+import tmp from 'tmp-sync';
+import WorkspaceAdapter from '../../lib/dependency-manager-adapters/workspace.js';
+import writeJSONFile from '../helpers/write-json-file.js';
+import assertFileContainsJSON from '../helpers/assert-file-contains-json.js';
+import generateMockRun from '../helpers/generate-mock-run.js';
+
+const fixtureWorkspaces = fs.readJsonSync('./test/fixtures/package-with-workspaces.json');
```
bertdeblock, line 10: Can be reverted to an import once we can use import attributes:

```js
import data from "https://example.com/data.json" with { type: "json" };
```

test/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
bertdeblock, line 1: Less work than renaming all files to `.mjs` and back again to `.js`.

PR #1036, opened 2024-12-18
Don't use ember-cli `project` instance outside of ember-cli commands

bertdeblock:
Part of decoupling `ember-try` from `ember-cli`.

This way, only the code in `lib/commands` is coupled to `ember-cli`.
The next step would be to move the `run` portion of the commands to separate files.
This would allow us to move everything to ES modules, except the commands.
The commands can async import and run the ES modules.

### DIFF
--- a/lib/utils/config.js
+++ b/lib/utils/config.js
@@ -5,12 +5,12 @@ const fs = require('fs');
 const { prefix, warn } = require('./console');
 const debug = require('debug')('ember-try:utils:config');
 
-function getConfigPath(project) {
+function getConfigPath(cwd) {
+  let packageFile = readPackageFile(cwd);
   let possibleConfigPath;
-  if (project.pkg && project.pkg['ember-addon'] && project.pkg['ember-addon']['configPath']) {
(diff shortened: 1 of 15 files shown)


bertdeblock:
Still need to update getting the config without a `project` instance.

kategengler (dismissed):

lib/utils/config.js
```diff
@@ -42,14 +42,13 @@ async function getBaseConfig(options) {
...
   let versionCompatibility =
-    options.versionCompatibility || versionCompatibilityFromPackageJSON(options.project.root);
+    options.versionCompatibility || versionCompatibilityFromPackageJSON(options.cwd);
   if (versionCompatibility) {
     // Required lazily to improve startup speed.
     let autoScenarioConfigForEmber = require('ember-try-config');
 
     let autoConfig = await autoScenarioConfigForEmber({
       versionCompatibility,
-      project: options.project,
```
bertdeblock, line 52: Does not seem to be used by `ember-try-config`?

kategengler (approved):

PR #1037, opened 2024-12-20
Remove passing an ember-cli `project` instance to the config function

bertdeblock:
I suspect (almost) no users will be affected by this.
If someone does run into this, feel free to give me a ping,
and I will try to suggest a proper migration path.


kategengler:
👍 It was also undocumented so this may not actually be breaking.

kategengler (approved):

PR #1038, opened 2024-12-20
Swap to pnpm

mansona:
With all the movement from @bertdeblock 's work I came here looking for a release-plan PR but we haven't set that up yet. I'm pretty sure that release-plan doesn't work with yarn classic so this PR is an attempt to swap over to pnpm. 

It's a bit tricky because of the dual package-manager implementation in the smoke-test-app but we'll see if CI is happy 😂


bertdeblock:
We can also set the `packageManager` field in the `package.json` file [like so](https://github.com/ember-cli/ember-cli/pull/10584).
Helps avoiding stuff [like this](https://github.com/ember-cli/ember-cli/pull/10583#discussion_r1890492668).
Also allows us to remove `with: version: 9` everywhere in the CI file,
so the package.json field is the single source of truth for the pnpm version.

bertdeblock (approved):
Ah, too bad!

PR #1039, opened 2024-12-20
Start using release-plan

mansona:


bertdeblock:
Is the `--single-package` option not needed anymore when running in a non-monorepo?

mansona:
@bertdeblock no it's not needed any more 👍 I actually have been meaning to remove it upstream 🙈 it's totally not needed any more

bertdeblock (approved):

PR #1063, opened 2025-01-29
[INTERNAL] Move logic out of `ember-cli` commands

bertdeblock:
After this, we can easily:

1. Migrate everything else (except `./lib/ember-cli-commands`) to `.mjs`
2. Update all our dependencies to latest because of `1.`
3. Introduce our own CLI to use `ember-try` without `ember-cli`


mansona:
what would it look like to invert this? i.e. make this a type=module package with an index.cjs ? I'm only asking because we're changing most files across to mjs and it seems like that is wanting to be type=module 🤔

bertdeblock:
I was under the assumption that importing cjs from inside a type=module package was not an option. But I assume this might work by configuring package.json#exports?

mansona:
yes you can configure package.json exports to point at a cjs file for the index 👍

PR #1085, opened 2025-03-06
Update release-plan workflows.

kategengler:
Used kategengler/create-release-plan#kg-reusable-actions (PR'd to create-release-plan)


kategengler:
I proposed the extraction to actions in create-release-plan-setup here https://github.com/embroider-build/create-release-plan-setup/pull/200

For now these actions are under one of my repos but they are tagged so there is no risk to using them as-is even before they are moved to a more shared repo.

NullVoxPopuli (approved):

PR #1089, opened 2025-03-08
Support ember-try.cjs

NullVoxPopuli:
in a type=module project, ember-try.js cannot use require, because js is an ESM


NullVoxPopuli:
I've stopped using ember-try, and I don't have time to follow up with this, so I'll close

